PR #448, opened 2021-04-10
Add single schedule query and remove schedule endpoints

gundamMC:
See notion for the new endpoints https://www.notion.so/berkeleytimeteam/Backend-V3-GraphQL-e632027417cc40dd9b8fb0f296558ef4#2964b2727203480a9c00c22d0543c810

Also fixes the create schedule error and the schedules query only querying my schedules lol


LEON-MING (approved):
sweet, looks good

PR #450, opened 2021-04-12
Civhacks promo + redirects

g-luo:
Create a RedirectLink object that opens the site in the url, ie berkeleytime.com/redirect?site=civhacks.
We open the url corresponding to this key in a map of allowedRedirects to prevent the possible attack where someone injects spam urls and has our website open them.


frontend/src/views/RedirectLink.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { FC } from 'react'
+import { useHistory, useLocation } from 'react-router-dom';
+
+// We must have an allowlist of redirects to prevent an attacker from arbitrarily opening external sites.
+const allowedRedirects = new Map([
+	["civhacks", "https://www.civhacks.com"],
+	["civhacks-register", "https://pzudnhexo5v.typeform.com/to/demEictZ"]
+])
```
hi-usui, line 8: This was a good call because I have tested exploits websites before that didn't have an allowlist for redirects like on typeracer.com haha

hi-usui:
logs for /civhacks and /civhacks-register working

Screen Shot 2021-04-12 at 16 05 44

hi-usui (approved):
core functionality is there
it's a little bit jittery when clicking on the links during redirect, but it's not a deal breaker since intermediate routes are appearing and that's what we need

PR #451, opened 2021-04-13
move test/ to statuspage/ since it's an actual server now

hi-usui:
did a lot of shifting, spiced up the syntax to be modern es6, made https://status.berkeleytime.com redirect to https://berkeleytime.statuspage.io via a 24/7 Express server listening on ports 80/443

TODO: somehow make the HTTPS certificates automatic

for future people trying to delete API key or see setup: `gsutil cp gs://berkeleytime-218606/secrets/statuspage.io -` for credentials to berkeleytime.statuspage.io


statuspage/index.js
```diff
@@ -0,0 +1,135 @@
...
+//     `berkeleytime.com/api/catalog/catalog_json/course_box/?course_id=2321`
+//   );
+// }, spConfig.restStatus);
+
+interval(async () => {
+  const { apiStatus, statusInterval, frontendStatus, url } = spConfig;
+  await updateStatuspage(async () => {
+    await axios.post(`${url}/api/graphql`, {
+      query: "query PingQuery{ ping }",
+    });
```
kevinzwang, line 94: I think I had some checks for the response status code as well as if the response data was valid JSON, cuz sometimes (at least the REST API) I think it defaults to sending some sort of HTML/XML when the backend is not working. Similar thing for the frontend.

hi-usui: The reason I wanted to remove those is because axios.post and axios.get are going to throw errors if the status isn't 200, so I believe those checks wouldn't be necessary, since you had already implemented catch logic on line 72-80

```
await axios.post
if (statusCode = 200 && response.data == "blah") {
... this block will never hit if statusCode is not 200 ...
... therefore only necessary if it's possible for status 200 to come back but still be failure case ...
}
```

kevinzwang: I see. sounds good to me then

kevinzwang (approved):

statuspage/index.js
hi-usui: image

the above test was for when I deleted the API key, but actually it's worth testing out a real backend failure then

hi-usui: @kevinzwang confirmed
image
image
image
image
image

PR #458, opened 2021-04-24
Update scheduler model and endpoints for public schedules

gundamMC:
Changes:

* Change total units to a string field, offloading calculations to frontend
* Change primary key to a uuid to improve security
* Add public field to schedule model and endpoints
* Add authentication requirement to user schema

For frontend: Most of the endpoints are detailed on Notion, but here's a quick overview:

* There is now a `totalUnits` string argument for creating and updating schedules that basically stores whatever you pass in as `totalUnits`
* There is also a `public` boolean argument for creating and updating schedules that determine whether of not a schedule is public (i.e. non-logged-in users and users other than the owner can view)
* Schedule models now use uuid, shouldn't affect too many things other than that their relay ids are a bit longer
* Queries for restricted fields will return null (e.g. you can't view another user's saved classes through their public schedule)

Note: Migrating a pk (esp. to uuid) when there are existing models is a bit awkward without manual migration. Since the existing ones are just testing schedules, the simplest way is to let django build the tables from scratch. I.e.:

1. Manually drop all the `scheduler_*` tables
2. Run `DELETE FROM django_migrations WHERE app = 'scheduler';` to remove django's migration history of scheduler
3. `python3 manage.py migrate scheduler`

I've already done this for staging but will prob need to do this on prod as well if the old migration was already migrated.


LEON-MING (approved):
niceeee

gundamMC:
btw i think we're good to merge this

PR #459, opened 2021-04-26
wrap each for-loop call with a try-catch block

hi-usui:
status.berkeleytime.com has gone down a few times as statuspage.io tells me to enhance our calm

instead of enhancing our calm, we wrap a try-catch that is catch-all

image


kevinzwang (approved):
LGTM

nociza:
Well thought implementation; could be the best solution to the problem.

PR #462, opened 2021-04-29
Grace 2021 cleanup

g-luo:
Please don't merge until May 1st.


g-luo:
TODO: add a check for the current date in components > ClassDescription > ClassDescription.tsx > checkOverridePrereqs. Let's also put this in a yaml file in public so we don't have to go into the code and modify.

PR #479, opened 2022-02-11
replace 0pm with 12pm

zacharyzollman:
It looks like the previous attempt to fix the noon label (https://github.com/asuc-octo/berkeleytime/pull/475) replaced 12am with 0pm. This is an attempt at fixing that.


hi-usui:
Thanks for the diligent follow-up!

Issue #490, opened 2022-03-27
Fix missing requirements


kevinzwang:
Ex:
- CS 61C should have the Physical Science breadth tag
- Cogsci 1 should have the Social and Behavioral Science breadth tag

We should find the root cause of the issue (SIS API not giving us all the data? Us dropping the data during preprocessing or serving?) and then develop a fix if possible

Boomaa23:
I've narrowed this down to some filter issue. We take in the data correctly from SIS (this is backed up by logs) and that data is written to the "temporary storage" CSV files correctly. The playlist supposedly updates, however none of the L&S requirements work (for any semester). Other filters work, just not the L&S ones. Will continue investigating.

Boomaa23:
Found it: see [here](https://github.com/asuc-octo/berkeleytime/blob/master/backend/playlist/schema.py#L18-L19)

The playlist queryset excludes all `ls` category playlists except for the current semester. Since the SIS API endpoint for the current semester (actually both Spring and Fall 2022) is returning a status of 401, the length of this playlist is zero and thus none of the L&S filters work. 

Regardless of this endpoint problem, it seems like excluding all but the most recent semester means none of the previous semester' L&S filters would work. I'm not as sure of that one though.

zacharyzollman:
That's interesting, it seems like filtering was introduced in [this commit](https://github.com/asuc-octo/berkeleytime/commit/a60c50961f98919c8ce80932f083c9deab6906b6), so maybe L&S requirements for past semesters have been missing since November 2020?

kevinzwang:
@Boomaa23 interesting. Could you find out why the endpoint is returning 401?

Boomaa23:
I believe this is because it uses the wrong API URL (this may have changed last year at some point). We haven't really seen the effects of this since it uses cached values for previous years (but as noted above these are not used anyways).

See [here](https://github.com/asuc-octo/berkeleytime/blob/master/backend/playlist/resource/ls.py#L40): The endpoint is: https://apis.berkeley.edu/sis/v1/classes/sections

When it should be (per [here](https://api-central.berkeley.edu/api/45/interactive-docs#/class%20section/getClassSectionsUsingGET)): https://apis.berkeley.edu/uat/sis/v1/classes/sections

Boomaa23:
Update: this comment:

[...]

PR #495, opened 2022-04-19
Fix backend L&S breadth parser endpoint URL

Boomaa23:
See #490, specifically this portion [here](https://github.com/asuc-octo/berkeleytime/issues/490#issuecomment-1101755470). Does not fix the static playlist selection issue.

### DIFF
--- a/backend/playlist/resource/ls.py
+++ b/backend/playlist/resource/ls.py
@@ -37,7 +37,7 @@ class LSResource(object):
         PAST_SEMESTERS_SIS + [{'semester': CURRENT_SEMESTER, 'year': CURRENT_YEAR}]
     )
     # Max page size is 100. We only want courses that are marked as active.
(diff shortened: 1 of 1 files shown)


hi-usui:
Hey, I think the cause might be misattributed? I tested both endpoints and both come up with the same payload when diffed.

We know for sure something causes a non-200 because of https://github.com/asuc-octo/berkeleytime/blob/master/backend/playlist/resource/ls.py#L150-L159, but the API requests preceding the 401 logs in 
[bt-backend-data-updater-staging.txt.gz](https://github.com/asuc-octo/berkeleytime/files/8516997/bt-backend-data-updater-staging.txt.gz) don't seem to be the ones causing the 401 errors

[uat.txt](https://github.com/asuc-octo/berkeleytime/files/8516973/uat.txt)
[sis.txt](https://github.com/asuc-octo/berkeleytime/files/8516974/sis.txt)

Boomaa23:
Yeah...seems to be something else. Will close this PR then.

PR #501, opened 2022-07-29
upd start/end dates to match 2022 Fall semester

Mehvix:
http://guide.berkeley.edu/academic-calendar/


christinas24 (changes requested):
- The months are zero-indexed, so should be 7 to 11.
- The last days of classes is Dec 2nd, so we may want to set the end to that. Dec 16th is the last day of finals.

Mehvix:
Ahhhh, thank you for catching those! I'll update the file right now.

christinas24 (approved):
Looks good, thanks for fixing this @Mehvix!

PR #506, opened 2022-08-24
dynamic calendar date generation

Mehvix:
My logic should be valid for [future calendars](https://registrar.berkeley.edu/calendar/#Current_and_Upcoming_Academic_Calendars). Implementing the various, overlapping summer sessions is a hassle with this date-dependent generation approach. 

Do the given low/high values seem reasonable (generate Fall schedules throughout May and up to November, Spring otherwise)?


christinas24 (changes requested):
Logic for generating the start/end dates looks good to me, but I think it'd be better to determine which semester/year it is from the `semester` parameter in line 210.

Changes requested:
- [ ] Determine fall/spring and year from `semester` param.
- [ ] Add some comments in the code explaining the logic for future developers.

PR #509, opened 2022-09-10
Disable recruitment banner

Gabe-Mitnick:
Comment out code that enables the recruitment banner. Commented out instead of deleted so that we can easily restore it for the next recruitment cycle.


kevinzwang (approved):
LGTM

PR #519, opened 2022-09-30
Update modal overflow

levikline:
Changed overflow from `scroll` to `hidden` in order to hide scroll bars on some devices.

### DIFF
--- a/frontend/src/assets/scss/bt/catalog/_modal.scss
+++ b/frontend/src/assets/scss/bt/catalog/_modal.scss
@@ -28,7 +28,7 @@
     .modal-content {
       border: 0px;
       border-radius: 0px;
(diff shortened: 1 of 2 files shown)


frontend/src/assets/scss/bt/catalog/_modal.scss
```diff
@@ -28,7 +28,7 @@
     .modal-content {
       border: 0px;
       border-radius: 0px;
-      overflow: scroll;
```
christinas24, line 31: I don't think we want to disable scroll here since this controls the modal on mobile views for the Catalog page.

christinas24 (changes requested):

Jdyn:
This has been fixed with the catalog rework

PR #522, opened 2022-10-01
No mobile scheduler

mathhulk:
Create a useMobile hook to determine whether the width of the document constitutes a mobile device and disable the Scheduler accordingly. Works with resize events.


mathhulk:
Will need to talk to the design team because the way the scheduler is currently designed/engineered makes responsiveness very difficult. Could use some quality of life updates.

christinas24 (approved):

PR #528, opened 2022-10-01
Improve appearance of grades graph on mobile

Gabe-Mitnick:
These were the problems that I had with the mobile version of the grades graph which I fixed

- square corners on bars instead of matching desktop version
- tooltip had weird thick border and square corners instead of matching desktop version
- when comparing classes, gap between bars within one grade bin was bigger than gab between different grade bins
- the numerical label on the widest bar would often get cut off
- the legend was at the bottom so I think mobile users would be confused until they scrolled all the way down (so I moved it to the top, but lmk if you disagree with that)
- grade axis would often only label every other category like A+, A-, B, C+, C-, even though there was space to just label all of them and make the graph less confusing
- one missing space in "Course Average A-(GPA: 3.709)" <- between the grade letter and the (GPA...)
- code formatting (I did the formatting in a separate commit so that you can select just the other commits to see the important diffs)

Before:
scrot 20
scrot 22

After:
scrot 18
scrot 21

Note, one weird thing that I didn't fix (because it ended up being harder than expected) is that, when comparing two classes, the tooltip on mobile only shows info for one class instead of for both. The mobile tooltip uses different code than desktop because it adds more info about percentiles to replace the info that's over to the right on desktop. On mobile, you can still see the info for each class by tapping on the different bars, but in a narrow desktop window, that doesn't work because of the way the bars respond to hovers, so you just can't get it to show the percentiles for one of the classes. I'm not sure if that bug/weirdness is worth fixing.

### DIFF
--- a/alumni.txt
+++ b/alumni.txt
@@ -1,1 +1,0 @@
-hiroshi '22 danji '22 vihan '24 annie '23 shuming '24
--- a/alumni.txt
+++ b/alumni.txt
@@ -1,0 +1,1 @@
+hiroshi '22 danji '22 vihan '24 annie '23 shuming '24
--- a/frontend/src/assets/scss/bt/grades/_grades.scss
+++ b/frontend/src/assets/scss/bt/grades/_grades.scss
@@ -25,7 +25,7 @@
   }
 
   .btn-primary:disabled {
-      cursor: default;
+    cursor: default;
   }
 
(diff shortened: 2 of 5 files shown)


christinas24 (dismissed):

christinas24 (approved):

kevinzwang (approved):
LGTM

PR #529, opened 2022-10-12
Added spring 2023 course catalog

williamT386:
Untested, since no environment file to use `make up` on. Will test with @kevinzwang


kevinzwang (approved):
LGTM

PR #551, opened 2023-02-02
Update graphql.ts

levikline:
Addresses the compile error:
`Type 'LazyQueryVariables' does not satisfy the constraint`


mathhulk (approved):
Need to add package-lock.json or remove ^ from versioning

PR #554, opened 2023-02-13
Add 2022 grade data

zacharyzollman:


backend/playlist/utils/config/abbreviation.yaml
```diff
@@ -262,6 +264,7 @@ ENV SCI:
 
 ETH STD:
     - "Ethnic Studies"
+    - "Chicanx Latinx Studies"
```
zacharyzollman, line 267: I think this should go under CHICANO. [Here](https://classes.berkeley.edu/content/2022-fall-chicano-39-001-sem-001) is an example of one of the courses that appears with this course subject.

backend/grades/resource/formatted/fall_2013.csv
```diff
@@ -1,2782 +1,2782 @@
-Abbreviation,CourseNum,CCN,SectionNum,CourseDesc,Instructor,Level,A+,A,A-,B+,B,B-,C+,C,C-,D+,D,D-,F,P,NP
```
zacharyzollman, line 1: I don't think we need to modify these csv files for other terms. If we do proceed with these modified versions, we should be sure to understand what exactly changed.

backend/grades/resource/formatted/fall_2022.csv
```diff
@@ -0,0 +1,3086 @@
...
+ETH STD,197,16370,014,Field Study,Luisa Giulianetti; Pablo Gonzalez; Andrew Henry,under,0,0,0,0,0,0,0,0,0,0,0,0,0,26,1
+ETH STD,197,16375,012,Field Study,Luisa Giulianetti; Pablo Gonzalez,under,0,0,0,0,0,0,0,0,0,0,0,0,0,11,1
+ETH STD,197,16380,005,Field Study,Victoria Robinson; Aileen Pagdanganan,under,0,0,0,0,0,0,0,0,0,0,0,0,0,26,1
+ETH STD,197,16384,009,Field Study,Laura Jimenez-Olvera; Dewey St Germaine; Keith Feldman,under,0,0,0,0,0,0,0,0,0,0,0,0,0,42,1
+ETH STD,197,16386,011,Field Study,Luisa Giulianetti; Pablo Gonzalez,under,0,0,0,0,0,0,0,0,0,0,0,0,0,21,1
+ETH STD,197,17376,001,Field Study,Laura Jimenez-Olvera; Pablo Gonzalez,under,0,0,0,0,0,0,0,0,0,0,0,0,0,27,1
+ETH STD,198,21052,008,Directed Group Stdy,Maria Gallegos-Diaz; Laura Jimenez-Olvera; Juan Berumen,under,0,0,0,0,0,0,0,0,0,0,0,0,0,26,1
+ETH STD,198,23302,008,Superv Group Study,Ronald Williams; Keith Feldman,under,0,0,0,0,0,0,0,0,0,0,0,0,0,17,2
+ETH STD,39,24421,001,Fresh/Soph Seminar,Christian Paiz,under,11,1,0,0,1,0,0,0,1,0,0,0,0,1,0
+ETH STD,39,33748,001,Fresh/Soph Seminar,Christian Paiz,under,16,0,0,0,0,0,0,0,0,0,0,0,0,1,0
```
zacharyzollman, line 730: In order for the change to the abbreviation to be reflected in the formatted csv, the script will need to be run again.

backend/grades/resource/formatted/fall_2013.csv
MichaelKhaykin: I am not sure that something did change - I think that all the files were overwritten so git thought I had basically deleted and re-added them - unless I am missing something while looking at the diffs.

MichaelKhaykin: Or actually maybe they did change because of the updated .yaml file - do we not want to keep those changes?

zacharyzollman: It does seem like the text is identical so it might be fine. The command for adding the data is specific to the term so unless we re-add data from past semesters, I don't see how this could cause any issues.

Jdyn: Looks like his editor changed the file line endings from CRLF to LF or vise versa. I would just stick to whatever it was originally imo.

kevinzwang (approved):
LGTM!

PR #556, opened 2023-02-19
Catalog Rework

Jdyn:
The catalog rework is open for initial testing and feedback

Below I list **only** changes which are noticable by the eye. I may make a comprehensive list of internal changes later. nothing is final:
- Added a design for usage with tablets
- reworked mobile filter design
- reworked 'class sections' design
- renamed 'reset' to 'clear' to clear the filters and aligned it right
- altered appearance of pills
- added a 'past offerings' section to quickly navigate to previous semesters of the course
- moved prerequisites outside of 'See more'
- Slightly redesigned Catalog course cards
    - Description can now span 2 lines to give more information
    - Grade is always displayed and aligned with course name and some other small changes
- Adding loading skeletons to course views to reduce cumulative layout shift
- Added accessible course search bar on mobile

This PR also contains the migration away from CRA and includes new ESLint, TS, and Prettier configs that offer sensible defaults that help catch bugs very well.

The rework can be viewed here: https://catalog.berkeleytime.com/


frontend/src/app/Catalog/CatalogList/CatalogList.module.scss
```diff
@@ -0,0 +1,145 @@
+.root {
+	display: flex;
+	grid-area: list;
+	gap: 10px;
+	position: relative;
+	flex-direction: column;
+	border-right: 1.5px solid #eaeaea;
+	height: calc(100% + 48px);
```
mathhulk, line 8: This height should be `100%` and not `calc(100% + 48px)`. A height greater than 100% causes the element to expand past the height of its parent before overflowing, which means the scrollbar gets cut off.

frontend/src/assets/scss/bt/base/_typography.scss
```diff
@@ -11,80 +11,92 @@ p,
...
+	color: $bt-base-text;
+	margin: 0;
 
-  -webkit-font-smoothing: antialiased;
-  -moz-osx-font-smoothing: grayscale;
+	-webkit-font-smoothing: antialiased;
+	-moz-osx-font-smoothing: grayscale;
+}
+
+::-webkit-scrollbar {
```
mathhulk, line 24: I would actually advise against altering the native appearance of the scrollbar. You will get different results depending on the operating system and browser and you will see unintended side effects unless you account for every edge case. For example, on Chrome on my Macbook, you now force scrollbars to show (when they don't normally) and you did not account for horizontal scrollbars, which have become large, clunky, and visually unappealing.

frontend/src/app/Catalog/CatalogFilters/CatalogFilters.module.scss
```diff
@@ -0,0 +1,195 @@
...
+		padding: 4px 8px;
+		border-radius: 4px;
+		font-size: 16px;
+		&:hover {
+			background: hsla(0, 0%, 92.2%, 0.6196078431);
+		}
+	}
+}
+
+.select {
```
mathhulk, line 169: You changed the design for the select inputs by increasing the font-size and font-weight and changing the placeholder values, which are no longer capitalized. You should change this back to match the current design.

frontend/src/app/Catalog/CatalogView/CatalogView.tsx
```diff
@@ -0,0 +1,247 @@
...
+								{req.name}
+							</span>
+						)
+					)}
+				</div>
+			</div>
+			<ReadMore>
+				{course.description !== '' ? course.description : 'This course has no description.'}
+			</ReadMore>
+			<h6>Prerequisites</h6>
```
mathhulk, line 187: Prerequisites should be within the `ReadMore` component as they are currently.

frontend/src/components/Recruiting/Description.jsx
```diff
@@ -1,58 +1,56 @@
-import React, { Component } from 'react';
```
mathhulk, line 1: Unless this file was referenced somewhere, you can remove it and the parent directory `Recruiting`.

mathhulk (approved):

PR #567, opened 2023-03-17
Update about carousel

QxBytes:
Jaden's path fixes merged in as well


frontend/src/lib/contributors.ts
```diff
@@ -23,88 +21,151 @@ export const current: Contributors = {
...
+				silly: path('henric_2.jpg')
 			}
 		},
 		{
-			name: 'Ethan Ikegami',
-			role: 'Backend Engineer',
-			site: 'https://ethanikegami.com/',
+			name: 'Jaden Moore',
+			role: 'Frontend Engineer',
+			site: 'https://www.jdyn.dev/',
```
Jdyn, line 67: Your version of this file deviates from the changes made in #566. #566 should be merged first and then you can remove the changes made to `contributors.ts` in this PR, assuming they don't pertain to the real changes you are attempting to make.

Jdyn (changes requested):

frontend/src/lib/contributors.ts
QxBytes: This version of the file should already include changes from #566 since I merged in that branch into this one and added some modifications (including missing past contributors etc.). I intend for the current `contributors.ts` in this branch with the changes it contains to be merged into master.

Jdyn: The merge wasn't encompassing because for example, I removed my site url, but in your merge, it was added back. I also reordered the json. imo the design team members should be moved up in the ordering.

QxBytes: Ah I see. Could you give me a list of changes you'd like and I can implement them? Alternatively feel free to commit to this branch with any changes you see fit. I would just keep the grouping of the Frontend/Backend/Design members together for consistency.

Jdyn (approved):
LGTM

PR #489, opened 2022-03-27
Add fall 2022 class config

Boomaa23:
It's that time again!

Follows [this](https://github.com/asuc-octo/berkeleytime/wiki/New-Semester-Runbook) guide.
Spring 2022 enrollment date information pulled from [here](https://chemistry.berkeley.edu/ugrad/current-students/important-dates).
Final exam group datetimes from [here](https://registrar.berkeley.edu/scheduling/academic-scheduling/academic-scheduling-final-exam-guide-and-schedules/).

Generally the same as #472 but for Fall 2022.

PR #573, opened 2023-03-30
Fall 2023 new semester updates

zacharyzollman:
- Did the normal stuff identified in the [associated wiki page](https://github.com/asuc-octo/berkeleytime/wiki/New-Semester-Runbook), as in PRs for [Spring 2023](https://github.com/asuc-octo/berkeleytime/pull/529), [Fall 2022](https://github.com/asuc-octo/berkeleytime/pull/489), [Spring 2022](https://github.com/asuc-octo/berkeleytime/pull/472), [Fall 2021](https://github.com/asuc-octo/berkeleytime/pull/453/files), *etc*.
- Also added abbreviations. Most of these were for [EAP](https://uceap.universityofcalifornia.edu/) courses, which showed up in data for Spring 2022, Fall 2022, and Spring 2023. Also added [Computational Precision Health](https://computationalhealth.berkeley.edu/), which showed up in Spring 2023.
- Also updated term start and end dates for exporting schedule to calendar. Found via [this PR](https://github.com/asuc-octo/berkeleytime/pull/501). We don't need to be hard coding these values; there is a PR to generate them dynamically [here](https://github.com/asuc-octo/berkeleytime/pull/506) and I think we should revisit that soon. CC @mathhulk on this last point

### DIFF
--- a/backend/berkeleytime/config/semesters/fall2023.py
+++ b/backend/berkeleytime/config/semesters/fall2023.py
@@ -8,7 +8,7 @@ CURRENT_YEAR = '2023'
 CURRENT_SEMESTER_DISPLAY = 'Fall 2023'
 
 # SIS API Keys
-SIS_TERM_ID = 2228
+SIS_TERM_ID = 2238
 
 TELEBEARS = {
     'phase1_start': datetime.datetime(2023, 4, 17),
--- a/backend/berkeleytime/config/semesters/fall2023.py
+++ b/backend/berkeleytime/config/semesters/fall2023.py
@@ -1,0 +1,42 @@
+"""Configurations for Fall 2023."""
+from berkeleytime.config.finals.semesters.fall2023 import *
+
+import datetime
+
+CURRENT_SEMESTER = 'fall'
+CURRENT_YEAR = '2023'
+CURRENT_SEMESTER_DISPLAY = 'Fall 2023'
+
+# SIS API Keys
+SIS_TERM_ID = 2228
+
+TELEBEARS = {
+    'phase1_start': datetime.datetime(2023, 4, 17),
+    'phase1_end': datetime.datetime(2023, 6, 16),
+    'phase2_start': datetime.datetime(2023, 7, 17),
+    'phase2_end': datetime.datetime(2023, 8, 13),
+
+    'adj_start': datetime.datetime(2023, 8, 14),
+}
+
+INSTRUCTION = {
+    'instruction_start': datetime.datetime(2023, 8, 23, 00, 00),
+    'instruction_end': datetime.datetime(2023, 12, 8, 00, 00)
+}
+
+# Please don't edit anything below this line unless you know what you are doing
+
+TELEBEARS_JSON = {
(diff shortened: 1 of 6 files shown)


backend/berkeleytime/config/semesters/fall2023.py
```diff
@@ -0,0 +1,42 @@
...
+    'phase1_end': datetime.datetime(2023, 6, 16),
+    'phase2_start': datetime.datetime(2023, 7, 17),
+    'phase2_end': datetime.datetime(2023, 8, 13),
+
+    'adj_start': datetime.datetime(2023, 8, 14),
+}
+
+INSTRUCTION = {
+    'instruction_start': datetime.datetime(2023, 8, 23, 00, 00),
+    'instruction_end': datetime.datetime(2023, 12, 8, 00, 00)
```
maxmwang, line 24: Should be `date time.datetime(2023, 12, 1, 00, 00)`. From [here](https://registrar.berkeley.edu/Calendar/), the last day of formal instruction is Dec 1, 2023. In previous years, the registrar marked the last day of RRR week (in this year: Dec 8, 2023) as the last day of instruction, but it makes more sense to indicate it as currently marked in the registrar because there is no instruction during RRR.

frontend/src/utils/scheduler/scheduler.ts
```diff
@@ -175,8 +175,8 @@ export const serializeSchedule = (schedule: Schedule, semester: Semester): Backe
  * Generates iCal file string.
  */
 export function scheduleToICal(schedule: Schedule, semester: Semester): string {
-	const SEMESTER_START = new Date(2023, 0, 17);
-	const LAST_COURSE_DAY = new Date(2023, 4, 6);
+	const SEMESTER_START = new Date(2023, 7, 23);
+	const LAST_COURSE_DAY = new Date(2023, 11, 8);
```
maxmwang, line 179: As noted before, `LAST_COURSE_DAY` should be `new Date(2023, 11, 1)`.

zacharyzollman: Oh yeah you're right, thanks for catching this.

backend/berkeleytime/config/semesters/fall2023.py
zacharyzollman: I think it's important to distinguish between the last day of *formal* instruction and the last day of (any) instruction. Some classes have review sessions, project presentations, or other instructional activities during RRR week, so RRR week is after the end of formal instruction but before the end of instruction. The last day of instruction isn't listed on the Google Calendar embedded on the registrar's page, but you can see it on the PDF version linked further down on the page. Based on the use of the last day of instruction (like, the end of RRR week) for `instruction_end` in previous terms, I believe it should refer to the end of (any) instruction. The date used for schedule export is different because the date is used for adding events to a calendar; regular class events don't generally occur during RRR week.

maxmwang: That's a good point, i didn't know there were still ongoing classes during RRR. Good catch

maxmwang (approved):

Issue #515, opened 2022-09-24
Improve search


Gabe-Mitnick:
The algorithm for course search on the catalog page should give better/more results and should be more performant. The relevant code is in `src/utils/search.ts` (or renamed to `search-utils.ts` in the `improve-search` branch) and `src/utils/courses/search.ts`

Examples of searches which should give better results:

- "algorithms" should surface COMPSCI 170 because its title contains "algorithms" (this is hopefully a simple bug, maybe
- "biology algorithms" should surface COMPSCI C176 and CMPBIO 276 instead of having to search "algorithms biology"
- Leave comments under this issue if you have more examples

Possibilities for improvement:

- Polish the `search()` algorithm, or replace it with a different more robust algorithm for comparing strings
    - Right now, I think it has a bug where it does `j++` in the condition of the `while` loop and looks at `j - 1` in the body where it should really look at `j - 2`
- Change the datastructure of `laymanToAbbreviation` to use something like a hashmap instead of an array of arrays

Gabe-Mitnick:
- "art8" has ART 8 as the highest result, but "art 8" with the space moves ART 8 further down
- "158" and "158A" should surface MUSIC 158A

PR #575, opened 2023-04-04
Improve Searching

### LINKED ISSUE #569: Weird Search Ranking
For some reason, when I tried searching for Chem 3b:
- if I typed *chem 3*
  - **CHEM 3B** appears first, then **CHEM 3A**
- if I typed *chem 3b*
  - **CHEM 3BL** appears first, then **CHEM 120B**, **CHEM 253B**, **CHEM 263B**, and then **CHEM 3B**

It doesn't really produce anything wrong, it's just the ranking is a bit confusing and sometimes require users to spend extra time to find their class when they are trying to be specific with their search string.

Jdyn:
This implementation takes inspiration from previous efforts in `improve-search` and `improve-search-2` detailed in #515.

From this we receive proper searchability in course dropdown menus like in the grades and enrollment page.
 - Typing `cs` properly results in `compsci` classes for example.
 - Typing course numbers like `16b` properly returns the relevent courses
 - Dropdown menus are implemented with a custom virtualization implementation to make them significantly faster to use and search, especially course dropdowns.
 - Catalog search is also improved from the changes.

Fixes #569 #515


mathhulk (approved):
Everything looks good to me!

PR #578, opened 2023-04-09
fix Inspirational Quote Retrieval™️ by adding QUOTES_API_KEY

hi-usui:
The API used for getting the quote of the day now requires authentication
Also, added the files which are slightly out of sync with changes at https://berkeleytime.com/git/berkeleytime/bt-gitlab-notify


hi-usui:
@kevinzwang Hey can we get this merged in? Doesn't affect things if we don't, but just in case as a precaution, want to avoid divergences from how things are running

kevinzwang (approved):
LGTM

PR #580, opened 2023-04-12
Enrollment resolver

Youngster-Joey:
Enrollment resolvers for the graphql API.
There are some desyncs between the data I have and the data used on the live website, idk where from. Need to investigate, as well as work on enrollment start/end dates and reworking the db.

If you have any questions about the code please lmk.

### DIFF
--- a/backend/src/modules/catalog/controller.ts
+++ b/backend/src/modules/catalog/controller.ts
@@ -267,11 +267,13 @@ export function getCourseList(): any {
 export async function getEnrollment(ccn: number, term: TermInput) {
     const section = await SectionModel.findOne({"id": ccn, "class.session.term.name": termToString(term)}).lean()
     const sectionObjectId = section?._id.toString()
-
-    // no idea why but the {$exists: true} part just does not work. If $exists works eventually, we should update the code to use it
-    // const diffs = await SectionHistoryModel.find({collectionId: new ObjectId(sectionObjectId), "diff.enrollmentStatus.enrolledCount": {$exists: true}}).lean()
-
+    
     const diffs = await SectionHistoryModel.find({collectionId: new ObjectId(sectionObjectId)}).lean()
+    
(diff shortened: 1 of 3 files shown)


backend/src/modules/catalog/controller.ts
```diff
@@ -262,3 +263,104 @@ export function getCourseList(): any {
         .sort({ "_id.subject": 1, "_id.number": 1 })
         .project({ _id: 0, subject: "$_id.subject", number: "$_id.number" })
 }
+
+export async function getEnrollment(ccn: number, term: TermInput) {
+    const section = await SectionModel.findOne({"id": ccn, "class.session.term.name": termToString(term)}).lean()
+    const sectionObjectId = section?._id.toString()
+
+    // no idea why but the {$exists: true} part just does not work. If $exists works eventually, we should update the code to use it
```
zacharyzollman, line 271: Let's try [`exists()`](https://mongoosejs.com/docs/api/query.html#Query.prototype.exists())

backend/src/modules/catalog/controller.ts
```diff
@@ -262,3 +263,104 @@ export function getCourseList(): any {
...
+    var prevEnrollCount = 0
+    var prevWaitlistCount = 0
+
+    var prevDay = firstDay
+
+    // interpolates missing dates
+    function fillGaps(currDay: Date, prevDay: Date, prevEnrollCount: number, prevWaitlistCount: number) {
+        let dayGap = daysElapsed(currDay, prevDay)
+
+        console.log("dg", dayGap)
```
zacharyzollman, line 317: Let's remove these (lines 317 and 319)

backend/src/modules/catalog/typedefs/catalog.ts
zacharyzollman, line 1: Let's add more docstrings to this file

zacharyzollman (changes requested):

PR #582, opened 2023-04-16
Add Catalog course name to browser tabs

Jdyn:
browser Tab title for the Catalog now updates to the specific course you are looking at.

This is implemented with a custom component so this can expand to all pages instead of just saying "Berkeleytime" everywhere.

This is also the first step toward proper discoverability on search engines.


Jdyn:
Adds browser metadata for the catalog which can easily expand to other pages in the future

Screenshot 2023-09-23 162730

Image of tab when viewing 61A in the catalog

mathhulk (approved):
LGTM

PR #589, opened 2023-08-26
fix invalid ical date generation

### LINKED ISSUE #587: Error while exporting to calendar
I am getting this error when trying to export to my calendar:

index-079878da.js:37 Uncaught RangeError: Invalid time value
    at Date.toISOString (<anonymous>)
    at n (SchedulerCalendar-53d23c9d.js:1:38863)
    at SchedulerCalendar-53d23c9d.js:9:36
    at Array.map (<anonymous>)
    at Wo (SchedulerCalendar-53d23c9d.js:1:39078)
    at x (Callout-ae4d6a55.js:1:24691)
    at Object.cA (index-079878da.js:37:9855)
    at dA (index-079878da.js:37:10009)
    at pA (index-079878da.js:37:10066)
    at ig (index-079878da.js:37:31466)

Jdyn:
There is currently a bug that prevents schedule export-to-calendar when any singular course within the schedule has a `startTime` and/or `endTime` that is null.

In these cases we basically take the time of whatever the first day of class function returns which appears reasonable and can simply be edited by the user after the export. Typically these start and end times were null for online classes that don't necessarily have a start and end time.

This fixes #587

### DIFF
--- a/frontend/src/utils/scheduler/scheduler.ts
+++ b/frontend/src/utils/scheduler/scheduler.ts
@@ -178,7 +178,19 @@ export function scheduleToICal(schedule: Schedule, semester: Semester): string {
 	const SEMESTER_START = new Date(2023, 7, 23);
 	const LAST_COURSE_DAY = new Date(2023, 11, 1);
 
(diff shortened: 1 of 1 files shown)


mathhulk (approved):
LGTM

PR #590, opened 2023-09-03
Auto-fill course selection on enrollment and grade URL navigation

Jdyn:
Feature requested by BT user:

> so a common flow is landing page -> explore courses -> look up class -> click average grade button but from there if u want to change it to focus on a certain semester/instructor u have to manually restart the class search at the top, like there's no direct way from here to, say, fall 2020 or smt. same issue holds for the enrollment page. i propose that the fields are either autofilled when u click the popup or there's an edit button added that lets u change the semesters/instructors for each box (like, next to the delete button)

This change updates the select components value when grade or enrollment data is said to be selected.

Changes viewable: https://auto-fill-course.berkeleytime.com/


Jdyn:
Since the components were rewritten to functional components, the implementation has become out of date. Should be very straight foward to work with FC. Maybe @xyntechx can if you have time.

PR #592, opened 2023-09-04
refactor to react-router v6

mathhulk:


frontend/src/components/Custom/H3.tsx
```diff
@@ -1,9 +1,6 @@
-import { FC } from 'react';
+import { PropsWithChildren } from 'react';
 import TextProps, { getClassNames } from './TextProps';
 
-const H3: FC<TextProps> = (props) => {
+export default function H3(props: PropsWithChildren<TextProps>) {
```
Jdyn, line 4: I do think propsWithChildren is nice in many cases where children is the only prop you are using and you don't want to create an entire type just for `children`. but in the case where you already have a type and are expecting many props, it is much less verbose to say `children: ReactNode;` instead of creating an import for `propsWithChildren` and then passing the current props as a generic etc.

Jdyn (changes requested):

frontend/src/app/Enrollment/index.jsx
```diff
@@ -0,0 +1,175 @@
...
+	enrollReset,
+	fetchEnrollFromUrl
+} from '../../redux/actions';
+import { useCallback, useEffect, useState } from 'react';
+
+const toUrlForm = (s) => {
+	return s.toLowerCase().split(' ').join('-');
+};
+
+export function Component() {
```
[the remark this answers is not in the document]
mathhulk: Understandable. There's probably a better way, but the `lazy` prop for routes requires a named export `Component`. See more: https://reactrouter.com/en/main/route/lazy. We can use `Component.displayName` to set the name of the component in React dev tools.

frontend/src/app/Grades/index.jsx
```diff
@@ -0,0 +1,168 @@
...
+		[navigate, selectedCourses]
+	);
+
+	const removeCourse = useCallback(
+		(id, color) => {
+			refillUrl(id);
+			dispatch(gradeRemoveCourse(id, color));
+		},
+		[dispatch, refillUrl]
+	);
```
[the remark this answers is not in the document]
mathhulk: Will change. For reference, I was working quickly and changed every function to `useCallback`.

frontend/src/components/Custom/H3.tsx
mathhulk: I think we should set a precedent here for using `PropsWithChildren` rather than defining the type yourself. You have to import a new type either way and I'd rather provide an example for incoming team members to not type a property manually that's already typed.

Jdyn: There is a significant problem with setting `PropsWithChildren` as a precedent. If you peak into the type you realize:
```tsx
type PropsWithChildren<P = unknown> = P & { children?: ReactNode | undefined };
```
This type forces children to be an optional type and disallows you to enforce the existence of children. For example If I create a modal component, I want to require that there is at least one child otherwise another developer might use my modal and not know what is happening and why nothing is rendering.

There is a greater issue though. it prevents the granular control we have of the form of `children`. for example, the current `ReadMore` component we are using, I ensure through types that the first child node must be a string, and the second child must be a `ReactNode` through `children: [string, ReactNode]`. This allows me to render text when its collapsed, and a react element when it's expanded.

In general, if I wanted to ensure that only two children are allowed for a component, I can enforce it: `children: [ReactNode, ReactNode];`. With `PropsWithChildren` I have no way of enforcing the form of the children, and further no way to require that the component have children in the first place.

frontend/src/app/Enrollment/index.jsx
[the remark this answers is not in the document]
Jdyn: Oh wow I didn't know about that requirement of `react-router` that is extremely strange of them to force for lazy loading. It could be a worthy sacrifice, but we could look into alternative lazy loading methods so we don't have to compromise.

frontend/src/components/Custom/H3.tsx
Jdyn: Also we can definitely enforce colocating the component prop types with the component itself like the catalog

```tsx
interface CatalogViewProps {
	coursePreview: CourseFragment | null;
	setCurrentCourse: Dispatch<SetStateAction<CourseFragment | null>>;
	setCurrentFilters: Dispatch<SetStateAction<CurrentFilters>>;
}

const skeleton = [...Array(8).keys()];

const CatalogView = (props: CatalogViewProps) => {
...
}
```
This way you always know where the component type is and don't need a separate file or import.

mathhulk: Very good point! We can change it where necessary.

Jdyn: Sounds good. Can you update the component types to reflect when children are required? Also are you still wanting to use `PropsWithChildren` in general? From my scan over the PR, it looks all these components require children to function

Issue #593, opened 2023-09-12
Add CORS header to allow localhost to access API


kevinzwang:
Add `localhost:3000` to the allowed origins for API requests, by setting the Access-Control-Allow-Origin headers

https://pypi.org/project/django-cors-headers/
https://github.com/asuc-octo/berkeleytime/blob/master/backend/berkeleytime/settings.py

PR #612, opened 2023-09-23
Use staging backend for frontend development

Jdyn:
Modify our api URLs to point toward the staging backend when we are in dev mode. Keep original functionality when in production.

Related #593 

Note intuitive change to use `npm run dev` instead of `start` because its shorter, more intuitive and standard now


mathhulk (approved):
lgtm

Issue #615, opened 2023-09-26
GraphQL Onboarding: Chat app


maxmwang:
Create a new complete module following our file structure. This module creates a chat app with 1 query and 1 mutation:

- query: find all messages, optionally filtering by user
- mutation: send a message containing user and message body. store this data in Mongo with a timestamp. 

Make sure to follow the file structure for our current modules! ie.:
- typedefs
- controller
- resolver
- formatter

Do this in a branch off of the `gql` branch, and submit a PR once you are done, tagging this issue and adding @maxmwang and @kevinzwang as reviewers

Jdyn:
This seems to be done. Feel free to reopen.

PR #617, opened 2023-09-30
added chat module

ARtheboss:
#615 
Hope this works


kevinzwang:
Will look at it soon! btw, do not add PRs to the projects page, but instead link them to the issue that is on the project. you should see a "Development" section the right sidebar where you can do that. I've done this for you this time

Issue #610, opened 2023-09-14
Remove application banner and redirect


kevinzwang:
- [x] /apply redirect
- [x] Application banner
- [x] Remove scheduler banner too

xyntechx:
PR: #627

PR #627, opened 2023-10-03
Remove application banner and redirect

xyntechx:
#610

### DIFF
--- a/frontend/src/components/Common/Layout.tsx
+++ b/frontend/src/components/Common/Layout.tsx
@@ -1,7 +1,7 @@
 import { Suspense, useEffect } from 'react';
 import { Outlet, useLocation } from 'react-router-dom';
 import ReactGA from 'react-ga';
-import Banner from './Banner';
+// import Banner from './Banner';
 import Navigation from './Navigation';
 import BTLoader from './BTLoader';
 import Footer from './Footer';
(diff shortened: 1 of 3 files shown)


frontend/src/components/Common/Layout.tsx
```diff
@@ -1,7 +1,6 @@
 import { Suspense, useEffect } from 'react';
 import { Outlet, useLocation } from 'react-router-dom';
 import ReactGA from 'react-ga';
-import Banner from './Banner';
```
Jdyn, line 4: Sorry there was some confusion, the intent was to prevent the banner from displaying in the DOM since we are no longer accepting applications, not removing the components entirely.

The way we've done it has just been to comment out, or remove the component from rendering.

Also I believe we still want the modal about the scheduler to appear still

Jdyn (changes requested):

frontend/src/components/Common/Layout.tsx
xyntechx: Got it!

About the scheduler modal, I checked with Matt and he said we should remove it. It's described in #610 as well :)

Jdyn: Ah sounds good!

xyntechx: Changes made :))

Jdyn (approved):
Nice.

PR #631, opened 2023-10-09
Jacky update django spring24

JackyJerkyYt:
update the website to spring 2024, as said on Monday, we don't need to use the final dates as we don't use them, so I didn't include them

### DIFF
--- a/backend/berkeleytime/config/semesters/spring2024.py
+++ b/backend/berkeleytime/config/semesters/spring2024.py
@@ -1,0 +1,41 @@
+"""Configurations for Fall 2023."""
+from berkeleytime.config.finals.semesters.spring2023 import *
+
+import datetime
+
+CURRENT_SEMESTER = 'spring'
+CURRENT_YEAR = '2024'
+CURRENT_SEMESTER_DISPLAY = 'Spring 2024'
+
+# SIS API Keys
+SIS_TERM_ID = 2242
+
+TELEBEARS = {
+    'phase1_start': datetime.datetime(2023, 10, 24),
+    'phase1_end': datetime.datetime(2023, 11, 5),
+    'phase2_start': datetime.datetime(2023, 11, 22),
+    'phase2_end': datetime.datetime(2023, 1, 7),
+    'adj_start': datetime.datetime(2023, 1, 8),
+}
+
+INSTRUCTION = {
+    'instruction_start': datetime.datetime(2024, 1, 16, 00, 00),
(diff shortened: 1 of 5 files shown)


backend/berkeleytime/config/semesters/spring2024.py
```diff
@@ -0,0 +1,41 @@
...
+
+# SIS API Keys
+SIS_TERM_ID = 2242
+
+TELEBEARS = {
+    'phase1_start': datetime.datetime(2023, 10, 24),
+    'phase1_end': datetime.datetime(2023, 11, 5),
+    'phase2_start': datetime.datetime(2023, 11, 22),
+    'phase2_end': datetime.datetime(2023, 1, 7),
+    'adj_start': datetime.datetime(2023, 1, 8),
```
maxmwang, line 18: Should be 2024.

maxmwang (changes requested):

backend/berkeleytime/config/semesters/spring2024.py
Youngster-Joey: also i think phase1 and phase2 start should be 10/16 and 11/14 respectively

frontend/src/utils/scheduler/scheduler.ts
```diff
@@ -175,8 +175,8 @@ export const serializeSchedule = (schedule: Schedule, semester: Semester): Backe
  * Generates iCal file string.
  */
 export function scheduleToICal(schedule: Schedule, semester: Semester): string {
-	const SEMESTER_START = new Date(2023, 7, 23);
-	const LAST_COURSE_DAY = new Date(2023, 11, 1);
+	const SEMESTER_START = new Date(2024, 1, 16);
+	const LAST_COURSE_DAY = new Date(2024, 4, 26);
```
kevinzwang, line 179: Javascript `Date` month values go from 0 to 11 (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date)

kevinzwang (changes requested):
Please include a finals config file as well. It may not be used in the frontend but it is depended on in [other parts of the code](https://github.com/asuc-octo/berkeleytime/blob/master/backend/catalog/mapper/section.py#L103). Just try your best to fill in the most accurate information.

kevinzwang (approved):
LGTM

maxmwang (approved):

PR #633, opened 2023-10-09
added authentication

Chengming-Li:


maxmwang (changes requested):
We want to make all requests regarding schedulers to require authentication, including mutations, except for quering by `id`, since we allow users to share schedules (if they're public). 

Current logic in Python backend is [here](https://github.com/asuc-octo/berkeleytime/blob/master/backend/scheduler/schema.py).

maxmwang (approved):

PR #662, opened 2023-11-28
finished implementing the redis for user session caching

JackyJerkyYt:
Haven't checked if the keys are really on the redis docker yet. But you can login now


maxmwang:
Resolved by #710.

PR #663, opened 2023-12-01
add browser title data for enrollment and grades

Jdyn:
Adds metadata for browser titles while navigating enrollment and grades

image


xyntechx (approved):
LGTM!

PR #686, opened 2024-03-13
fr-releases

mathhulk:


frontend/src/App.tsx
```diff
@@ -52,6 +53,15 @@ const router = createBrowserRouter([
       },
     ],
   },
+  {
+    element: <Layout />,
```
mathhulk, line 57: You can put the `/releases` path within the existing layout on line 39.

PR #687, opened 2024-03-13
fr-about

mathhulk:


frontend/src/app/About/index.tsx
```diff
@@ -1,3 +1,194 @@
...
+        reader.readAsDataURL(blob);
+      });
+    } catch (error) {
+      console.error('Error converting image to Base64:', error);
+      throw error;
+    }
+  }
+
+  //get data from airtable
+  function airtableDB() {
```
mathhulk, line 84: It might be better practice to separate these functions from your component and into a `.ts` file. I would try to keep as much logic outside of the React component as possible (to a reasonable extent).

frontend/package.json
```diff
@@ -17,13 +17,19 @@
 		"prod": "serve -s build -p 3000"
 	},
 	"dependencies": {
+		"airtable": "^0.12.2",
+		"bt": "^0.0.1",
```
mathhulk, line 21: Remember to remove this dependency.

frontend/src/app/About/About.module.scss
```diff
@@ -0,0 +1,315 @@
+//undefined stuff that was in main branch
+$bt-navbar-height: 58px;
+$bt-light-text: #8A8A8A;
+$bt-base-text: #383838;
```
mathhulk, line 4: We have Tailwind colors defined as CSS variables in the `src/main.scss` file.

You can reference CSS variables like so:
```css
color: var(--slate-500);
```

frontend/src/app/About/Contributors.tsx
```diff
@@ -0,0 +1,96 @@
...
+		console.log(alumniContributors)
+		console.log(currContributors[0]?.name)
+		console.log(currContributors[0]?.img.seriousBase64);
+	
+	 */
+
+	console.log(currContributors)
+	const base64Root = 'data:image/jpeg;base64,'
+
+	const alumniByGradYear: { [key: number]: contributorStructure[] } = {};
```
mathhulk, line 22: Not super important, but you can convert this and the following `forEach` loop into a `useMemo` React hook using the `[Array.prototype.reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce)` function. Memoization will prevent re-renders if they were to occur.

frontend/src/app/About/Contributors.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { contributorStructure } from './index';
+import styles from "./Contributors.module.scss"
+
+interface ContributorsProps {
+	currContributors: contributorStructure[];
+	alumniContributors: contributorStructure[];
+}
+
+
+const Contributors: React.FC<ContributorsProps> = ({ currContributors, alumniContributors }) => {
```
mathhulk, line 11: When you are writing component, I would always use syntax like:
```ts
export default function Contributors({ currContributors, ... }: ContributorsProps) {
   ...
}
```
When you export anonymous functions as components, they aren't properly handled by React developer tools and you need to define `Contributors.displayName = "Contributors"` before you export, but I find it much easier to export as I showed above.

frontend/src/app/About/Contributors.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+
+	return (
+		<>
+			<div className={styles.currentContributors}>
+				<h1>
+					Current Contributors
+				</h1>
+				<div>
+					{currContributors.map(({ name, img, websiteURL, role }) => (
```
mathhulk, line 43: It might be easier to include these loops within `About/index.tsx` and instead separate `Contributor` and `Alumni` into their own reusable components that you pass `name, img, websiteURL, etc.` into from the loop.

PR #688, opened 2024-03-13
basic faq layout, no accordion

mathhulk:


frontend/src/App.tsx
```diff
@@ -52,6 +53,15 @@ const router = createBrowserRouter([
       },
     ],
   },
+  {
+    element: <Layout footer={false} />,
+    children: [
+      {
+        element: <FAQ />,
+        path: "/faq",
```
mathhulk, line 61: You can put the `/faq` path into the existing layout route defined on line 39. Routes can have an unlimited number of children, so you can just add the FAQ next to the About child.

PR #708, opened 2024-06-14
grades FERPA compliance

maxmwang:
- exclude classes with <=10 total graded
- exclude classes with only 1 grade bin (ie. 100% students got the same grade)


mathhulk (approved):
best PR i've ever seen

PR #710, opened 2024-06-16
use redis to store user session cookies

maxmwang:
Adapted from branch [user_redis_cache](https://github.com/asuc-octo/berkeleytime/tree/user_redis_cache) worked on by @JackyJerkyYt. 

Stores user session cookies in Redis instead of in memory (default behavior). After deploying, backend instances can be brought back up from 1.


kevinzwang (approved):
LGTM, feel free to merge if this is tested

PR #711, opened 2024-06-16
Merge frontend work into GQL branch

mathhulk:
- Modify `catalog` query to return `Course` and `Class` interfaces rather than `CatalogCourse` and `CatalogClass` in order to enable client-side filtering
- Type all the enums to view possible values in the GraphQL explorer


backend/src/bootstrap/loaders/passport.ts
```diff
@@ -77,36 +125,39 @@ export default async (app: Application) => {
...
-    if (!email) {
-      return done(null, false, { message: 'No email found' });
-    }
+  passport.use(
+    new GoogleStrategy.Strategy(
+      {
+        clientID: config.GOOGLE_CLIENT_ID,
+        clientSecret: config.GOOGLE_CLIENT_SECRET,
+        callbackURL: LOGIN_REDIRECT,
+      },
```
maxmwang, line 134: Are `scope` and `state` not necessary here?

maxmwang (approved):

backend/src/modules/catalog/typedefs/catalog.ts
```diff
@@ -1,142 +1,319 @@
...
 
     """
-    Get info about all courses and their corresponding classes for a given semester. 
-    
+    Get info about all courses and their corresponding classes for a given semester.
+
     Used primarily in the catalog page.
     """
-    catalog(term: TermInput!): [CatalogItem]
+    catalog(term: TermInput!): [Course!]
```
kevinzwang, line 25: The reason why I had a separate item from `Course` here originally was to ensure that it was performant enough. The catalog query was already quite slow, although proper indexing may help that a lot, but just be cautious about this, especially if a random person tries to make a nested query (e.g. `{ catalog { classes { sections { number } } } }`), that could really put a load on the server and/or database

kevinzwang (approved):
I'm not qualified to review the frontend code and I don't think I have enough context to understand the changes to the passport loader, but I am approving this PR on the understanding that both of those are properly written and tested. 

Had one comment but otherwise changes to catalog look good to me, glad to see additions to the API schema for data the frontend is finding that they need!

backend/src/modules/catalog/typedefs/catalog.ts
maxmwang: @mathhulk might have more to say but he said he's planning to add a max depth on the queries.

kevinzwang: If so please make an issue or something for that and make sure it gets done before this is released to the public!

PR #733, opened 2024-10-27
feat: updater-scripts

adit-bala:


apps/datapuller/src/shared/parser.ts
```diff
@@ -0,0 +1,191 @@
+import { ISectionItem } from "@repo/common";
+import { ClassSection } from "@repo/sis-api/classes";
+
+function getRequiredField<T>(
+  value: T | undefined,
+  fieldName: string,
+  defaultValue: T
+): T {
+  if (value === undefined || value === null) {
+    console.warn(`Missing required field: ${fieldName}`);
```
maxmwang, line 10: Should probably use the `log` object. We can pass this down from the as a parameter (as opposed to use a global instance). 

This might be easiest to do by placing the `log` object into the `config` object, and then passing the `config` object downstream.

apps/datapuller/src/shared/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+      console.log(`Error fetching page ${page} of data`);
+      console.log(`Unexpected error querying API. Error: "${error}"`);
+
+      if (retries === 0) {
+        console.log(`Too many errors querying API. Terminating update...`);
+        break;
+      }
+
+      retries--;
+      console.log(`Retrying...`);
```
maxmwang, line 55: Let's change this `console.*` to use the `log` object as well.

apps/datapuller/src/class.ts
```diff
@@ -0,0 +1,52 @@
...
+import mapClassToNewClass, { CombinedClass } from "./shared/classParser";
+import { Config } from "./shared/config";
+import { fetchActiveTerms, fetchPaginatedData } from "./shared/utils";
+
+async function updateClasses(config: Config) {
+  const log = config.log;
+  const classesAPI = new ClassesAPI();
+
+  log.info("Fetching Active Terms");
+  const activeTerms = await fetchActiveTerms(log, {
```
mathhulk, line 14: For updating classes and sections, I would recommend fetching both `Current` and `Next` terms, as future terms are likely to change quite often as well.

apps/datapuller/src/course.ts
```diff
@@ -1,43 +1,52 @@
...
+import { fetchActiveTerms, fetchPaginatedData } from "./shared/utils";
 
-async function main() {
-  const { log } = setup();
+async function updateCourses(config: Config) {
+  const log = config.log;
+  const coursesAPI = new CoursesAPI();
+
+  log.info("Fetching Active Terms");
+  const activeTerms = await fetchActiveTerms(log, {
```
mathhulk, line 14: Courses aren't related to terms, so you don't have to fetch terms in this pipeline

apps/datapuller/src/section.ts
```diff
@@ -0,0 +1,52 @@
...
+import { Config } from "./shared/config";
+import mapSectionToNewSection from "./shared/sectionParser";
+import { fetchActiveTerms, fetchPaginatedData } from "./shared/utils";
+
+async function updateSections(config: Config) {
+  const log = config.log;
+  const classesAPI = new ClassesAPI();
+
+  log.info("Fetching Active Terms");
+  const activeTerms = await fetchActiveTerms(log, {
```
mathhulk, line 14: For updating classes and sections, I would recommend fetching both `Current` and `Next` terms, as future terms are likely to change quite often as well.

apps/datapuller/src/class.ts
adit-bala: I believe this is already done in `utils.ts`

apps/datapuller/src/section.ts
adit-bala: duplicate comment

apps/datapuller/src/section.ts
```diff
@@ -0,0 +1,67 @@
...
+    },
+    (data) => data.apiResponse.response.classSections || [],
+    mapSectionToNewSection
+  );
+
+  log.info("Example Section:", sections[0]);
+
+  log.info(`Updated ${sections.length} sections for Spring 2024`);
+
+  await SectionModel.deleteMany({});
```
mathhulk, line 38: You should only delete sections you are going to add. This should look like:
```ts
await SectionModel.deleteMany({
  "class.session.term.id": { "$in": activeTerms }
});
```
(if `activeTerms` is an array of term ids)

apps/datapuller/src/class.ts
```diff
@@ -0,0 +1,67 @@
...
+      app_id: config.sis.CLASS_APP_ID,
+      app_key: config.sis.CLASS_APP_KEY,
+    },
+    (data) => data.apiResponse.response.classes || [],
+    mapClassToNewClass
+  );
+
+  log.info("Example Class:", classes[0]);
+
+  await ClassModel.deleteMany({});
```
mathhulk, line 36: You should only delete classes you are going to add. This should look like:
```ts
await ClassModel.deleteMany({
  "session.term.id": { "$in": activeTerms }
});
```
(if `activeTerms` is an array of term ids)

apps/datapuller/src/class.ts
```diff
@@ -0,0 +1,69 @@
...
+  );
+
+  log.info("Example Class:", classes[0]);
+
+  await NewClassModel.deleteMany({
+    "session.term.id": { $in: activeTerms },
+  });
+
+  // Insert classes in batches of 5000
+  const insertBatchSize = 5000;
```
mathhulk, line 41: Just leaving this here, but we should probably do some testing on `hozer` to determine the final batch size. It all depends on amount of memory, etc.

apps/datapuller/src/section.ts
```diff
@@ -0,0 +1,67 @@
...
+      app_id: config.sis.CLASS_APP_ID,
+      app_key: config.sis.CLASS_APP_KEY,
+    },
+    (data) => data.apiResponse.response.classSections || [],
+    mapSectionToNewSection
+  );
+
+  log.info("Example Section:", sections[0]);
+
+  await NewSectionModel.deleteMany({});
```
mathhulk, line 36: You still need to add filters to this (only delete what you're adding)

apps/datapuller/src/shared/classParser.ts
```diff
@@ -0,0 +1,55 @@
...
+  )?.id;
+  const termId = original.session?.term?.id;
+  const sessionId = original.session?.id;
+  const number = original.number;
+
+  if (!courseId || !termId || !sessionId || !number) {
+    throw new Error("Missing essential class fields");
+  }
+
+  const newClass: IClassItem = {
```
mathhulk, line 34: You're missing `courseNumber` and `subject` it seems, and `year` and `semester`

maxmwang (approved):

PR #734, opened 2024-10-30
Monitor crowd-sourced data pod

mathhulk:


apps/backend/src/modules/rating/controller.ts
```diff
@@ -0,0 +1,334 @@
...
+  const aggregated = await ratingAggregator(classIdentifier);
+  if (!aggregated.length) return null;
+
+  return formatAggregatedRatings(aggregated[0]);
+};
+
+export const createRating = async (
+  context: any, 
+  ratingIdentifier: RatingIdentifier,
+  email: string,
```
mathhulk, line 31: You don't need to take in `email` because you can get the user from the `getUser(context)` function in the `user` module.

apps/backend/src/modules/rating/typedefs/rating.ts
```diff
@@ -0,0 +1,123 @@
...
+    Usefulness
+    Difficulty
+    Workload
+
+    "Boolean metrics"
+    Attendance
+    Recording
+}
+
+enum Semester {
```
mathhulk, line 15: You shouldn't have to redefine this variable since `Semester` has already been defined in the `term` module.

apps/backend/src/modules/rating/typedefs/rating.ts
```diff
@@ -0,0 +1,123 @@
...
+"""
+Ratings by class
+"""
+type AggregatedRatings {
+    "Class identifer"
+    subject: String!
+    courseNumber: String!
+    semester: Semester!
+    year: Int!
+    class: String!
```
mathhulk, line 31: To keep naming consistent, I would use `number` or `classNumber` to represent a class here.

apps/backend/src/modules/rating/typedefs/rating.ts
```diff
@@ -0,0 +1,123 @@
...
+"""
+Get data
+"""
+input ClassIdentifier {
+    "Class Identifiers"
+    subject: String!
+    courseNumber: String!
+    semester: Semester!
+    year: Int!
+    class: String!
```
mathhulk, line 82: Same for `class` here. Use `number` or `classNumber`.

apps/backend/src/modules/rating/typedefs/rating.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+type Query {
+    aggregatedRatings(
+        classIdentifier: ClassIdentifier!
```
mathhulk, line 86: I would suggest flattening this input and writing it as `aggregatedRatings(subject: String!, courseNumber: string!, ...)` so as to not make overly-nested queries.

apps/backend/src/modules/rating/typedefs/rating.ts
```diff
@@ -0,0 +1,123 @@
...
+        isAllTime: Boolean!
+    ): AggregatedRatings!
+
+    userRatings: UserRatings! @auth
+
+    semestersOffered(
```
mathhulk, line 92: You also don't need this query since this data is already available on the `course` module.

apps/backend/src/modules/rating/typedefs/rating.ts
```diff
@@ -0,0 +1,123 @@
...
+    courseNumber: String!
+    semester: Semester!
+    year: Int!
+    class: String!
+
+    metricName: MetricName!
+}
+type Mutation {
+    createRating(
+        rating: RatingIdentifier!
```
mathhulk, line 113: I would definitely recommend spreading the input here rather than doing a nested input, just like I said before in the query.

apps/frontend/src/App.tsx
```diff
@@ -143,6 +148,10 @@ const router = createBrowserRouter([
             element: <Class.Grades />,
             path: "grades",
           },
+          {
+            element: <CatalogRatings />,
```
mathhulk, line 152: You can see how the components are defined as `<Class.Component />` which is what you should have here as it is part of the larger `<Class />` component:
```ts
element: <Class.Ratings />
```

apps/frontend/src/components/Class/Overview/index.tsx
```diff
@@ -1,24 +1,21 @@
...
-}
+    const { class: _class } = useClass();
+    return (
+        <div className={styles.root}>
+            <Details {..._class.primarySection.meetings[0]} />
+            <p className={styles.label}>Description</p>
+            <p className={styles.description}>
+                {_class.description ?? _class.course.description}
+            </p>
+            <AttendanceRequirements
```
mathhulk, line 15: I noticed you removed the requirements. Was this on purpose?

apps/frontend/src/components/UserFeedbackModal/RatingForm.tsx
```diff
@@ -0,0 +1,76 @@
...
+  });
+
+  const handleRatingClick = (type: keyof typeof ratings, value: number) => {
+    setRatings(prev => ({
+      ...prev,
+      [type]: prev[type] === value ? 0 : value
+    }));
+  };
+
+  const renderRatingScale = (
```
mathhulk, line 23: I would recommend splitting this into a separate component within this file rather than using a render function, as I personally find render functions a bit less structured and readable.

apps/frontend/src/components/UserFeedbackModal/RatingForm.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <span>{rightLabel}</span>
+      </div>
+    </div>
+  );
+
+  return (
+    <div className={styles.ratingSection}>
+      <h2 className={styles.sectionTitle}>Course Ratings</h2>
+      
+      {renderRatingScale(
```
mathhulk, line 54:
```jsx
<RatingScale type={...} question={...} ... />
```

apps/frontend/src/components/UserFeedbackModal/UserFeedbackModal.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
```
mathhulk, line 1: You shouldn't (and won't need to) import `React` itself anywhere. Instead, only import the definitions you need.

apps/frontend/src/components/UserFeedbackModal/UserFeedbackModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+  subtitle, 
+  currentClass 
+}: UserFeedbackModalProps) {
+  return (
+    <Dialog.Root open={isOpen} onOpenChange={onClose}>
+      <Dialog.Portal>
+        <Dialog.Overlay className={styles.overlay} />
+        <Dialog.Content className={styles.modal}>
+        <Dialog.Close className={styles.closeButton}>
+            ✕
```
mathhulk, line 30: You can use [Iconoir icons](https://iconoir.com/docs/packages/iconoir-react):
```jsx
<Xmark />
```

apps/frontend/src/components/Class/Overview/index.tsx
```diff
@@ -1,24 +1,212 @@
...
+  Yes = "Yes",
+  No = "No",
+  BellowThreshold = "BellowThreshold",
+  Indeterminate = "Indeterminate",
+}
+
 export default function Overview() {
   const { class: _class } = useClass();
+  return (
+    <CourseContext.Provider value={{ course: _class.course }}>
```
mathhulk, line 38: The `useClass` already has a `course` field, so the additional `CourseContext` here is unnecessary.

mathhulk, line 38: The `useClass` already has a `course` field, so the additional `CourseContext` here is unnecessary.

apps/frontend/src/components/Class/Overview/index.tsx
```diff
@@ -1,24 +1,212 @@
...
+        User-Submitted Class Requirements
       </p>
-      {_class.course.requirements && (
-        <>
-          <p className={styles.label}>Prerequisites</p>
-          <p className={styles.description}>{_class.course.requirements}</p>
-        </>
-      )}
+      <div>
+        {(() => {
```
mathhulk, line 94: I'd stay away from this paradigm and instead:
```ts
            {response.Attendance === Consensus.Yes && (
                <>
                  <User className={styles.icon} />
                  <span className={styles.userSubmissionDescription}>
                    Attendance Required
                  </span>
                </>
              )}
            {response.Attendance === Consensus.No && (
                <>
                  <UserXmark className={styles.icon} />
                  <span className={styles.userSubmissionDescription}>
                    Attendance Not Required
                  </span>
                </>
              )}
```

apps/frontend/src/components/Class/Overview/index.tsx
```diff
@@ -1,24 +1,212 @@
...
+                  <span className={styles.userSubmissionDescription}>
+                    Attendance Not Required
+                  </span>
+                </>
+              );
+          }
+        })()}
+      </div>
+      <div>
+        {(() => {
```
mathhulk, line 118: This auto-executing functions are really an anti-pattern here. Same with the switch statements. You have no default for them and you lose efficiency (though that's not really my biggest concern) and readability. I'd recommend the typical `{conditional && <>...</>}` instead

apps/frontend/src/components/Class/Overview/index.tsx
```diff
@@ -1,24 +1,212 @@
...
+        className={styles.suggestEdit}
+        onClick={handleFeedbackClick}
+      >
+        Look inaccurate? Suggest an edit →{" "}
+      </Link>
     </div>
   );
 }
+
+function getResponse(course: ICourse, metricName: MetricName): Consensus {
```
mathhulk, line 186: If helper functions are used regularly, I'd place them in a common file in `lib` or in your `@repo/shared` package

apps/frontend/src/components/Class/index.tsx
```diff
@@ -136,8 +137,8 @@ export default function Class({
   const [updateUser] = useUpdateUser();
 
   const { data: course, loading: courseLoading } = useReadCourse(
-    providedClass?.subject ?? (subject as string),
-    providedClass?.courseNumber ?? (courseNumber as string)
+    providedClass?.subject || (subject as string),
```
mathhulk, line 140: These were `??` because `??` is only truthy for `null` and `undefined` values. `||` is truthy for all truthy values like empty strings and 0. Changing this unnecessary reduces the strictness of this check

apps/frontend/src/components/Class/index.tsx
```diff
@@ -224,6 +224,16 @@ export default function Class({
     );
   }, [_class, bookmarked, updateUser, user]);
 
+  function getRatingsCount() {
+    if (!_class?.course?.aggregatedRatings?.metrics?.length) {
+      return 0;
+    }
+    return _class.course.aggregatedRatings.metrics.reduce(
+      (max: number, metric: any) => Math.max(max, metric.count),
```
mathhulk, line 232: `any` should not be used as a type ever

apps/frontend/src/components/Class/index.tsx
```diff
@@ -402,31 +423,51 @@ export default function Class({
                   <MenuItem active={isActive}>Grades</MenuItem>
                 )}
               </NavLink>
+              <NavLink to={{ pathname: "ratings" }}>
+                {({ isActive }) => (
+                  <MenuItem active={isActive}>
+                    Ratings
+                    {(() => {
+                      const count = getRatingsCount();
```
mathhulk, line 431: Same as I said before

packages/shared/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@repo/shared",
+  "version": "0.0.0",
+  "private": true,
+  "main": "index.ts",
+  "types": "index.ts",
```
mathhulk, line 6: You can remove this `types` field as the package will never be built and will always be evaluated by the parent

packages/theme/src/components/Tooltip/index.tsx
```diff
@@ -57,3 +56,42 @@ export const Tooltip = forwardRef<
     );
   }
 );
+
+export const AlternateTooltip = forwardRef<
```
mathhulk, line 60: If you want to provide an alternate tooltip, I would specifically define a new component folder and name it relevant to when this component should be used. The components in `@repo/ui` are meant specifically as building blocks and not to satisfy *every* possible use case. I would move this to the frontend repo most likely!

apps/frontend/src/lib/api/courses.ts
```diff
@@ -30,6 +30,20 @@ export interface ICourse {
   title: string;
   toDate: string;
   typicallyOffered: Semester[] | null;
+  aggregatedRatings?: {
+    __typename?: "AggregatedRatings";
```
mathhulk, line 34: I would recommend typing `aggregatedRatings` and `metrics` within the `ratings.ts` file and importing it here since it is used in multiple places

apps/frontend/src/lib/api/classes.ts
```diff
@@ -181,6 +183,28 @@ export interface IMeeting {
   instructors: IInstructor[];
 }
 
+export interface IMetric {
```
mathhulk, line 186: Move to `ratings.ts`!

apps/frontend/src/components/Class/Ratings/RatingsContainer.tsx
```diff
@@ -0,0 +1,418 @@
+import React, { useEffect, useState } from "react";
```
mathhulk, line 1: There's no need to import `React` itself anywhere! Just import the functions or types you need

apps/frontend/src/components/Class/Ratings/RatingsContainer.tsx
```diff
@@ -0,0 +1,418 @@
...
+
```
mathhulk, line 2: I would also just move this file to the `index.tsx` since there's no need to just have the `index.ts` export this file

PR #742, opened 2024-11-18
Monitor GradTrak

mathhulk:


packages/common/src/models/gradtrak.ts
```diff
@@ -0,0 +1,202 @@
...
+export const SelectedCourseModel = mongoose.model<SelectedCourseType>('course', selectedCourseSchema);
+
+export type MajorReqType = Document & InferSchemaType<typeof majorReqSchema>;
+export const MajorReqModel = mongoose.model<MajorReqType>('majorReq', majorReqSchema);
+
+export interface PlanTermType extends Document {
+  name?: string;
+  courses: SelectedCourseType[];
+  user_email: string;
+  term?: {
```
mathhulk, line 184: I would suggest storing the `year` and `semester` flat here rather than nested within `term` as `year` and `planTerm`. We use this same standard throughout the rest of the collections.

packages/common/src/models/gradtrak.ts
```diff
@@ -0,0 +1,202 @@
...
+  },
+  is_minor: {
+    type: Boolean,
+    required: false
+  }
+});
+
+export const gradtrakSchema = new Schema(
+  {
+    user_email: {
```
mathhulk, line 140: I would stick with a standard for casing, and I'd suggest `camelCase` instead of `snake_case`.

packages/common/src/models/gradtrak.ts
```diff
@@ -0,0 +1,202 @@
...
+  },
+  college_reqs: {
+    type: [String],
+    required: false,
+    trim: true,
+  }
+});
+
+export const selectedCourseSchema = new Schema({
+  class_ID: {
```
mathhulk, line 50: If you are storing classes, then you should probably reference the schedule schema. If you are only storing courses (which is probably the only thing necessary to start), then you only need to store the `courseId` and not any of the class or section identifiers.

packages/common/src/models/gradtrak.ts
```diff
@@ -0,0 +1,202 @@
...
+    required: true,
+    trim: true,
+    alias: "end",
+  },
+  title: {
+    type: String,
+    required: false,
+    alias: "name",
+  },
+  location: {
```
mathhulk, line 21: I would assume some of this additional information (`location`, `start_time`, `end_time`, `days_of_week`) probably isn't required for GradTrak because it doesn't deal with data of that specificity.

packages/common/src/models/gradtrak.ts
```diff
@@ -0,0 +1,202 @@
...
+  user_email: string;
+  planTerms: PlanTermType[];
+  miscellaneous: PlanTermType;
+  uni_reqs: string[];
+  college_reqs: string[];
+  major_reqs: MajorReqType[];
+  createdAt: Date;
+  updatedAt: Date;
+}
+export const GradtrakModel = mongoose.model<GradtrakType>('gradtrak', gradtrakSchema);
```
mathhulk, line 202: I would also suggest naming this `PlanModel` instead of `Gradtrak` to keep the terminology generic.

apps/backend/src/modules/gradtrak/typedefs/gradtrak.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    """
+    Edits the list of major requirements
+    """
+    editMajorRequirements(major_reqs: [MajorReqInput!]!): Gradtrak @auth
+
+    """
+    Deletes gradtrak, for testing purposes
+    """
+    deleteGradtrak: String @auth
```
mathhulk, line 231: I would suggest naming this `Term` and not `Gradtrak` like I was talking about before.

apps/backend/src/modules/gradtrak/typedefs/gradtrak.ts
```diff
@@ -0,0 +1,235 @@
...
+  type Mutation {
+    """
+    Takes in user's email, creates a new Gradtrak record in the database, and returns the Gradtrak
+    """
+    createNewGradtrak: Gradtrak @auth
+
+    """
+    Takes in a new college, switches the user's college requirements
+    """
+    changeGradtrakCollege(college: Colleges!): Gradtrak @auth
```
mathhulk, line 199: I would suggest switching to a generic `updatePlan` that takes in a new college and any other plan-specific fields rather than an endpoint for every action.

PR #743, opened 2024-11-20
(cicd) install new helm charts when deploying

maxmwang:
Before:
- On `hozer`, there is a clone of the repo at branch `gql`. The deploy step would `cd ./infra` and `helm install app` from there. This means that helm chart modifications are not included in the CICD pipeline, requireing a manual `git pull` or `git switch` to deploy new helm charts. 

After:
- On `hozer`, local clone of repo is not necessary. The deploy step installs a version of the helm chart from docker hub (same registry used for container images). This means helm modifications can be deployed in dev/stage environments without affecting the prod environment. 

TODO:
- Only the `app` chart is installed per deployment. There are three other possibly relevant charts: `base`, `mongo`, and `redis`. Realistically, these charts should not be deployed in dev/stage environments. Fortunately, changes to these charts will be small and infrequent, so manually installations of these charts upon updates is okay. 
- Helm chart versioning does not use the recommended [semver](https://semver.org/) convention. This isn't fully necessary (our container images don't either), but would be nice for built in rollback with `helm upgrade --atomic`.


klhftco (approved):
nice

PR #747, opened 2024-11-21
(core) integrate semantic search pod

mathhulk:


maxmwang:
This PR should eventually include the weaviate datapuller

PR #759, opened 2025-01-09
Basic Profile Page

ARtheboss:
- Awaiting some changes from catalog and crowd sourced data branch to be merged before this is finalized


apps/frontend/src/components/NavigationBar/index.tsx
```diff
@@ -54,13 +54,20 @@ export default function NavigationBar({ invert }: NavigationBarProps) {
             </MenuItem>
           )}
         </NavLink>
+        <NavLink to="/profile">
+          {({ isActive }) => (
+            <MenuItem className={styles.item} active={isActive}>
+              Profile
```
mathhulk, line 60: Instead of adding a new tab for the profile, I'd recommend combining the sign in button with the profile button since log out is not actually a primary action. A log out button can instead be placed within the profile (or account) page.

PR #760, opened 2025-01-15
Catalog dashboard page updates

ARtheboss:
Need help checking two things: 

- Switch terms on the catalog dashboard page and bookmark on class page in catalog are disabled. For some reason, react `disable={false}` causes the disabled attribute to still be there on my local machine (with value "false"), when it should be removed.
- Sign in to add bookmark, verify it appears on Dashboard

Will integrate "View all" bookmark button with profile once both these branches are merged


apps/frontend/src/components/CarouselClass/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  );
+
+  const _class = useMemo(() => data, [data]);
+
+  return loading || !_class ? (
+    <></>
+  ) : (
+    <div
+      className={styles.root}
+      onClick={() => onSelect(subject, courseNumber, number)}
```
mathhulk, line 45: Instead of using an `onClick` event, you can just use a link:
```tsx
<Link to={`/catalog/${year}/${semester}/${subject}/${courseNumber}/${number}`}>
  ...
</Link>
```

You could also instead use a `ClassDialog` for minimal impact on navigation:
```tsx
<ClassDialog
  year={year}
  semester={semester}
  subject={subject}
  courseNumber={courseNumber}
  number={number}
>
  ...
</ClassDialog>
```

apps/frontend/src/app/Catalog/Dashboard/index.tsx
```diff
@@ -3,36 +3,126 @@ import { Dispatch, SetStateAction } from "react";
...
+                        (term) =>
+                          term.semester === semester && term.year === year
+                      )
+                  )
+                  .map((t) => {
+                    return (
+                      <DropdownMenu.Item
+                        key={`${t.semester} ${t.year}`}
+                        onClick={() => {
+                          changeTerm(t);
```
mathhulk, line 113: Instead of passing down a `changeTerm` function from a new state variable, you should just be using a link. The URL in the catalog should be considered the source of truth for what the user intends to see:
```ts
onClick={() => navigate(`/catalog/${t.year}/${t.semester}`)}
```

You also significantly reduce the amount of data you have to pass around and store.

apps/frontend/src/app/Catalog/Dashboard/index.tsx
```diff
@@ -45,23 +135,79 @@ export default function Dashboard({
...
-          <div className={styles.card}></div>
-          <div className={styles.card}></div>
-        </Carousel>
-        <Carousel title="Upcoming events" Icon={<Calendar />} to="/semesters">
+        <p className={styles.heading}>
+          {term.semester} {term.year}
+        </p>
+        {term.startDate && term.endDate ? (
+          <p className={styles.paragraph}>
+            {formatDate(new Date(term.startDate))} through{" "}
```
mathhulk, line 143: Instead of using a custom `formatDate` function, I've already installed and imported [Moment.js](https://momentjs.com/) elsewhere that you can use:
```ts
{moment(term.startDate).format("MMMM Do")}
```

apps/frontend/src/app/Catalog/Dashboard/index.tsx
```diff
@@ -45,23 +135,79 @@ export default function Dashboard({
...
+                      courseNumber={c.courseNumber}
+                      number={c.number}
+                    />
+                  </div>
+                );
+              })
+          )}
         </Carousel>
+        {recents.length == 0 ? (
+          <></>
```
mathhulk, line 186: You can remove unnecessary fragments by using a logical AND:
```ts
{recents.length > 0 && (...)}
```

apps/frontend/src/components/Class/index.tsx
```diff
@@ -232,6 +234,14 @@ export default function Class({
   // TODO: Error state
   if (!course || !_class || !pin) {
     return <></>;
+  } else {
+    addToRecent({
```
mathhulk, line 238: Right now, `addToRecent` will be called on every render. You should instead use a `useEffect` for this:
```ts
useEffect(() => {
    if (!course || !_class) return;

    addToRecent({
      subject: _class.subject,
      year: _class.year,
      semester: _class.semester,
      courseNumber: _class.courseNumber,
      number: _class.number,
    });
}, [course, _class]);
```

apps/frontend/src/components/Class/index.tsx
```diff
@@ -254,7 +264,7 @@ export default function Class({
                     [styles.active]: bookmarked,
                   })}
                   onClick={() => bookmark()}
-                  disabled={userLoading}
+                  disabled={userLoading} // setting disabled to false still appears on my version
```
mathhulk, line 267: This is a bug I will address!

apps/frontend/src/lib/recents.ts
```diff
@@ -0,0 +1,45 @@
...
+    (v, rcd) =>
+      v ||
+      (newRcd.subject == rcd.subject &&
+        newRcd.year == rcd.year &&
+        newRcd.semester == rcd.semester &&
+        newRcd.courseNumber == rcd.courseNumber &&
+        newRcd.number == rcd.number),
+    false
+  );
+  if (alreadyExists) return;
```
mathhulk, line 36: It might be nice to move the class to the front of the array if it already exists.

apps/frontend/src/components/Class/index.tsx
ARtheboss: There's a similar bug with the switch term select on the Dashboard component. It sets `data-disabed="false"` which disables the select

apps/frontend/src/components/CarouselClass/index.tsx
ARtheboss: I couldn't find a `ClassDialog` component, so using `Link` instead

PR #762, opened 2025-01-20
Schedules

mathhulk:
- [x] Share dialog 
  - Restricted to public (anyone with the link) vs. private for now
  - Pave the way for sharing with individual users in the future and possibly collaboration
- [x] Add, edit, and remove custom events
- [x] Duplicate (copy) schedule
  - [x] Toggle to disable custom events
- [x] Edit name
- [x] Select schedule to compare
---
- [x] Create dialog
  - [x] Select term
- [x] List schedules
  - [x] Display recently viewed schedules using persistent storage


mathhulk:
Work still needs to be done to finalize many of the features and iterate on the design, but I'm merging this for now.

PR #774, opened 2025-02-02
Gql courses pins

ARtheboss:
- Some basic design for course page
 - Basic design for pins page. Was thinking in the future can do a collapsible menu by term.


mathhulk:
We’ll leave this for later so I will keep this as a draft for now.

PR #786, opened 2025-02-21
complete modules rewrite

maxmwang:
## Backend

Refactors all resolvers/controllers to use new database models. Query designs are kept as identical as possible. 

Small GraphQL typdef changes including:
- `sessionId` with all queries requiring year/semester
  - field isn't required, default value is '1' (explicit at each query). 
  - if semester is 'Summer', will silently return [] (unless valid `sessionId` is provided)
- section type now includes latest `EnrollmentSingular`, including status (replacing `open`)
- scheduler auto-populates sectionIds as sections objects

Renamed `classNumber` field to `sectionId` field in grade-distribution schema. 

## Frontend 
Refactors frontend variable naming:
- enrollCount -> enrolledCount
- enrollMax -> maxEnroll
- waitlistCount -> waitlistedCount
- waitlistMax -> maxWaitlist

This reflects the SIS API naming. 

Also, added an API file for the enrollment queries. 

## TODO:
- [ ] delete all old models
- [ ] rename new models (remove "New" prefixes)
- [ ] deploy to prod and create indices (not auto created)

### DIFF
--- a/apps/datapuller/src/lib/grade-distributions.ts
+++ b/apps/datapuller/src/lib/grade-distributions.ts
@@ -47,7 +47,7 @@ const formatDistribution = (distribution: GradeDistributionRow) => {
     termId: distribution.semester_year_term_cd,
     sessionId: distribution.session_code,
 
-    classNumber: distribution.class_number,
+    sectionId: distribution.class_number, // not sure why tf the api is like this
     sectionNumber: distribution.class_section_cd,
 
     count: parseInt(distribution.grade_count),
--- a/apps/frontend/src/components/ClassBrowser/browser.ts
+++ b/apps/frontend/src/components/ClassBrowser/browser.ts
@@ -59,7 +59,7 @@ export const getFilteredClasses = (
       // Filter by open
       if (
         currentOpen &&
-        _class.primarySection.enrollment.latestEnrollment.status !== "O"
+        _class.primarySection.enrollment.latest.status !== "O"
       ) {
         acc.excludedClasses.push(_class);
 
--- a/apps/frontend/src/components/ClassBrowser/browser.ts
+++ b/apps/frontend/src/components/ClassBrowser/browser.ts
@@ -57,7 +57,10 @@ export const getFilteredClasses = (
   return classes.reduce(
     (acc, _class) => {
       // Filter by open
-      if (currentOpen && !_class.primarySection.open) {
+      if (
+        currentOpen &&
+        _class.primarySection.enrollment.latestEnrollment.status !== "O"
+      ) {
         acc.excludedClasses.push(_class);
 
         return acc;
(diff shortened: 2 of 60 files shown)


packages/common/src/models/sectionNew.ts
```diff
@@ -24,20 +24,6 @@ export interface ISectionItem {
   primary?: boolean;
   type?: string;
   combinedSections?: number[];
-  enrollment?: {
```
mathhulk, line 27: Why was enrollment removed?

apps/frontend/src/lib/api/terms.ts
```diff
@@ -42,7 +42,7 @@ export const READ_TERMS = gql`
       sessions {
         name
         startDate
-        endDate
```
mathhulk, line 45: Did you mean to duplicate startDate?

packages/common/src/models/classNew.ts
```diff
@@ -72,18 +72,18 @@ const classSchema = new Schema<IClassItem>({
   requirementDesignation: { type: String }, // NOTE: Exclude if always the same as course requirementsFulfilled, requirementDesignation.code
   requisites: { type: String }, // requisites.description
 });
-// classes are unique by termId, courseId, number, subject, and courseNumber
 classSchema.index(
   {
-    termId: 1,
+    year: 1,
+    semetser: 1,
```
mathhulk, line 78: Spelling error

apps/frontend/src/lib/api/schedules.ts
```diff
@@ -167,26 +178,31 @@ export const UPDATE_SCHEDULE = gql`
...
             startDate
             endDate
-            ccn
             component
-            enrollCount
-            enrollMax
-            waitlistCount
-            waitlistMax
+            enrollment {
+              latestEnrollment {
```
mathhulk, line 198: Hmm this is an interesting paradigm

apps/frontend/src/components/ClassBrowser/browser.ts
```diff
@@ -57,7 +57,10 @@ export const getFilteredClasses = (
   return classes.reduce(
     (acc, _class) => {
       // Filter by open
-      if (currentOpen && !_class.primarySection.open) {
+      if (
+        currentOpen &&
+        _class.primarySection.enrollment.latestEnrollment.status !== "O"
```
mathhulk, line 62: Do you think there might be a better place for status rather than nested within `enrollment.latestEnrollment`. Maybe we can chat about this today. I'm curious about your thoughts here

apps/datapuller/src/lib/grade-distributions.ts
```diff
@@ -47,7 +47,7 @@ const formatDistribution = (distribution: GradeDistributionRow) => {
     termId: distribution.semester_year_term_cd,
     sessionId: distribution.session_code,
 
-    classNumber: distribution.class_number,
+    sectionId: distribution.class_number, // not sure why tf the api is like this
```
mathhulk, line 50: `class_number` also represents CCN which I believe stands for Common Course Numbering, which is a blanket system implemented by colleges across the US, but I think they just do a really bad job of differentiating in their systems at Berkeley

PR #799, opened 2025-03-02
Rachel

racheler15:
finished backend onboarding


packages/common/src/models/course.ts
```diff
@@ -1,5 +1,29 @@
...
+  createdAt: string;
+}
+
+// creates a schema and tells mongo how a comment should be stored
+const CommentSchema = new Schema<CommentItem>({
+  subject: { type: String, required: true },
+  number: { type: String, required: true },
+  userEmail: { type: String, required: true },
+  content: { type: String, required: true },
+  createdAt: { type: String, default: () => new Date().toISOString() },
```
mathhulk, line 18: Mongoose actually supports timestamp columns (`updatedAt` and `createdAt`) out of the box: https://mongoosejs.com/docs/timestamps.html

packages/common/src/models/course.ts
```diff
@@ -1,5 +1,29 @@
 import { Document, Model, Schema, model } from "mongoose";
 
+// typescript interface for a comment
+export interface CommentItem {
+  subject: string;
+  number: string;
+  userEmail: string;
```
mathhulk, line 7: Only nit I would have here is using `userId` or a relationship instead of `userEmail`: https://mongoosejs.com/docs/populate.html

apps/backend/src/modules/course/resolver.ts
```diff
@@ -52,6 +54,24 @@ const resolvers: CourseModule.Resolvers = {
 
       return courses as unknown as CourseModule.Course[];
     },
+
+    comments: async (_, { subject, number, userEmail }, _context, _info) => {
```
mathhulk, line 58: I think it might have been better to create a new module instead of adding on to the course module! Then, you can later add a reference resolver on top of the existing course query to include comments for that course too!

packages/common/src/models/course.ts
```diff
@@ -1,5 +1,29 @@
 import { Document, Model, Schema, model } from "mongoose";
 
+// typescript interface for a comment
+export interface CommentItem {
```
mathhulk, line 4: You also should include an ID:
```ts
_id: Schema.Types.ObjectId
```

apps/backend/src/modules/course/formatter.ts
```diff
@@ -29,3 +29,13 @@ export function formatCourse(course: ICourseItem) {
 
   return output;
 }
+
+export function formatComment(message: CommentItem) {
+  return {
+    subject: message.subject,
+    number: message.number,
```
mathhulk, line 36: Nothing is really wrong here, but since you are returning all of the fields anyways, you can just:
```ts
return message;
```

apps/backend/src/modules/course/resolver.ts
```diff
@@ -52,6 +54,24 @@ const resolvers: CourseModule.Resolvers = {
...
+      const allComments = await getComment(
+        subject,
+        number,
+        userEmail || undefined
+      );
+      return allComments.map(
```
mathhulk, line 64: Because you've already created a format function, use it!
```ts
return allComments.map(formatComment);
```

PR #800, opened 2025-03-05
Rachel2

racheler15:


apps/frontend/src/components/Class/Comments/Comments.module.scss
```diff
@@ -0,0 +1,38 @@
+.root {
+  --foreground-color: #000; /* Default for light mode */
```
mathhulk, line 2: These variables are already defined for you, so you don't have to add them

apps/frontend/src/components/Class/Comments/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <label htmlFor="text-input">Make a new comment:</label>
+      <input
+        id="text-input"
+        type="text"
+        value={value}
+        onChange={(e) => setValue(e.target.value)}
+        className={styles.input}
+        placeholder="Say anything..."
+      />
+      <button onClick={handleSubmit} className={styles.button}>
```
mathhulk, line 52: Try using one of the reusable components from `@repo/theme`:
```ts
import { Button } from "@repo/theme";

<Button onClick={() => handleSubmit()}>Submit</Button>
```

PR #812, opened 2025-04-08
change restore prod from cronjob to job

maxmwang:


maxmwang:
We're keeping cronjob

PR #813, opened 2025-04-09
improved bookmarking responsiveness + saved classes and schedules in user profile

Youngster-Joey:
fix: improved catalog carousels' responsiveness when bookmarking pages from the main catalog screen
feat: added name, bookmarked classes, and saved schedules to user profile page


apps/frontend/src/components/Course/index.tsx
```diff
@@ -180,7 +180,7 @@ export default function Course({
         optimisticResponse: {
           updateUser: {
             ...user,
-            bookmarkedCourses,
+            bookmarkedCourses: user.bookmarkedCourses,
```
mathhulk, line 183: Shouldn't this reference `bookmarkedCourses` and not `user.bookmarkedCourses`?

apps/frontend/src/components/Class/index.tsx
```diff
@@ -208,7 +208,7 @@ export default function Class({
         optimisticResponse: {
           updateUser: {
             ...user,
-            bookmarkedClasses,
+            bookmarkedClasses: user.bookmarkedClasses,
```
mathhulk, line 211: Same here?

ARtheboss (approved):

PR #827, opened 2025-08-01
Change maxage from 1 hour to 24 hours

quincythai:


ARtheboss:
Ideally, we should be using refresh tokens. @subhashjprasad maybe this is something profile can look into.

PR #835, opened 2025-09-05
Meetings fix

ARtheboss:
"Section date interpretation bug" and "Scheduler/Catalog multi lecture location only shows one lecture bug" Notion tickets fix


Chengming-Li (approved):
LGTM

PR #848, opened 2025-09-30
Reimplement to use catalog search for gradtrak

Chengming-Li:


ARtheboss:
@Chengming-Li Looks like the add classes flow is really slow and annoying to use. Any way we can hide this latency?

ARtheboss (approved):

PR #849, opened 2025-09-30
labels

ARtheboss:


Chengming-Li:
@ARtheboss bug: deleting a label does not remove it from existing classes, but removes it from the menus. This means if you delete a label, that label is now permanently stuck on all classes with that label before deletion

Chengming-Li (approved):

Chengming-Li:
@ARtheboss fixed the bug and changed the adding tag system to not use the button, take a look

PR #851, opened 2025-10-02
Implement add semester functionality

Chengming-Li:
Merging into labels branch because I branched off that branch

### DIFF
--- a/apps/backend/src/modules/plan/controller.ts
+++ b/apps/backend/src/modules/plan/controller.ts
@@ -13,6 +13,34 @@ import {
 } from "../../generated-types/graphql";
 import { formatPlanTerm, formatPlan } from "./formatter";
 
+// Helper functions for chronological insertion
+function getTermOrder(term: string): number {
+  switch (term) {
+    case 'Spring': return 1;
+    case 'Summer': return 2;
+    case 'Fall': return 3;
+    default: return 0;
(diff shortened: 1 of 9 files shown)


Chengming-Li:
> @Chengming-Li Whenever I add a semester, it shows that it contains the same classes as Fall 2022 (my first semester). When I add another one, the old one I added no longer shows the Fall 2022 classes but now the new one does. If I reload, everything is correct tho so this seems to imply that there is something wrong with the local state?

For the sake of repro, are you creating a duplicate semester?

ARtheboss:
> We can backlog this for now I think cuz they're currently sorted in chronological order

Hmm because then hide semester doesn't work either, because how you unhide it is you drag it out of the hidden list.

ARtheboss:
> For the sake of repro, are you creating a duplicate semester?

I have a duplicate semester but the duplication is also happening with non-duplicate semesters.

ARtheboss:
I'm lowkey gonna merge this and fix it in the semester block details code I'm writing

Chengming-Li:
> I'm lowkey gonna merge this and fix it in the semester block details code I'm writing

Okay, I haven't had a chance to look into it but my hunch is that each planterm list item is using its name as the key, so that's causing the duplications. Changing it to its id should work

PR #853, opened 2025-10-03
Fix bug with deleting tags

Chengming-Li:
Issue: if you delete a tag and then add it back, all the classes that had the tag will have the tag again. This is probably fine.


Chengming-Li:
> > Issue: if you delete a tag and then add it back, all the classes that had the tag will have the tag again. This is probably fine.
> 
> 
> 
> This is a feature not a bug. Is that the issue getting fixed here?

No, that's something I just noticed. This PR fixes a bug where if you delete a tag, the frontent content reverts to the initial tags when you first loaded the dashboard.

PR #862, opened 2025-10-07
Fix bug with custom class not removing placeholder

Chengming-Li:


ARtheboss:
Getting this error: `index.tsx:385 Error handling drop: TypeError: allSemesters[sourceSemesterId] is not iterable
    at handleDrop (index.tsx:358:43)`

PR #867, opened 2025-10-08
Fix grade distribution

PineND:
Fix bug + made stuff look a bit better


ARtheboss (approved):
LGTM assuming it works lol. Good idea to put session id along with term info

PR #880, opened 2025-10-14
(cicd) auth functionality in dev deploy

maxmwang:
Previously, auth was not supported in dev deploy due to the commit hash-based subdomains that we host dev deployments to. These dynamically generated subdomains is not easily matched in the Google OAuth configs. 

Now, upon passing in a boolean flag during the build step, the subdomains can be fixed (`auth0.dev.stanfurdtime.com`, ..., `auth9.dev.stanfurdtime.com`). 

> [!WARNING] 
> If you deploy a dev branch without the auth flag, then try to deploy _the same commit_ with the auth flag, it will cause an error. To resolve this, you have to ssh into the machine and `helm uninstall` the original release. (This behavior existed even before this PR unfortunately: https://www.notion.so/berkeleytimeteam/fix-cicd-dev-bug-1d8a86d7e7df8073a920c1cf1805d04b?source=copy_link)

### DIFF
--- a/.github/workflows/cd-dev.yaml
+++ b/.github/workflows/cd-dev.yaml
@@ -19,7 +19,7 @@ on:
         default: false
 
 env:
-  HOSTS: "beta.berkeleytime.com auth1.stanfurdtime.com"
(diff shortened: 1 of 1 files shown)


ARtheboss (approved):

maxmwang:
> > If you deploy a dev branch without the auth flag, then try to deploy _the same commit_ with the auth flag, it will cause an error
> 
> Why would this still happen? It seems like a different auth host would be calculated so there shouldn't be any conflict, right?

Yeah, the problem has nothing to do with the auth domains (this bug has existed even before this PR). The problem is some immutability problem with a k8s resource that is being updated on the second deploy.

PR #882, opened 2025-10-14
Catalog Requirements Filter

aaureliaw:


apps/frontend/src/components/ClassBrowser/browser.ts
```diff
@@ -45,14 +48,58 @@ export const getLevel = (academicCareer: AcademicCareer, number: string) => {
     : (academicCareers[academicCareer] as Level);
 };
 
+export const getBreadthRequirements = (sectionAttributes: any[]): Breadth[] => {
```
ARtheboss: Don't think this argument needs to be of type `any`

ARtheboss:
image
From the API docs, seems like we also have the `active`, `fromDate`, and `toDate` fields in APICentral. I think these should also be included. Seems like they might tell the story of when these classes actually meet these requirements, but if you've seen from the data that they're useless, this is probably fine too.

apps/datapuller/src/lib/sections.ts
```diff
@@ -106,6 +106,18 @@ const formatSection = (input: ClassSection) => {
...
+        code: attr.attribute?.code || "",
+        description: attr.attribute?.description || "",
+        formalDescription: attr.attribute?.formalDescription || "",
+      },
+      value: {
+        code: attr.value?.code || "",
+        description: attr.value?.description || "",
+        formalDescription: attr.value?.formalDescription || "",
+      },
+    })),
```
ARtheboss: requirementDesignation should also be changed to be like this. Right now it's just a code which isn't readable.

aaureliaw:
i ran this mongodb query to see if any section had active set to false and it returned empty so idt there are inactive GE requirements 

Screenshot 2025-10-19 at 5 50 04 PM

PR #884, opened 2025-10-15
Catalog pod

PineND:


thompsonngo:
LGTM! @NathanDai5287

PineND:
LGTM

apps/backend/src/modules/enrollment/typedefs/enrollment.ts
```diff
@@ -39,15 +39,21 @@ export default gql`
...
     maxWaitlist: Int!
-    openReserved: Int
+    openReserved: Int!
     instructorAddConsentRequired: Boolean
     instructorDropConsentRequired: Boolean
-    seatReservationCounts: [SeatReservationCount!]
+    seatReservationCounts: [SeatReservationCounts!]
+  }
+
+  type SeatReservationCounts {
```
ARtheboss: Shouldn't be plural

apps/backend/src/modules/enrollment/typedefs/enrollment.ts
```diff
@@ -12,7 +12,7 @@ export default gql`
     ): Enrollment
   }
 
-  type Enrollment @cacheControl(maxAge: 300) {
+  type Enrollment {
```
ARtheboss: So this max age is why it keeps refreshing the cache. Maybe it should be changed to like 15min * 60 = 900 for now? Or we can just keep as is, but we would need to put in the correct caching behavior like ASAP (otherwise enrollment numbers on catalog will be off)

apps/frontend/src/components/AverageGrade/index.tsx
```diff
@@ -74,7 +153,12 @@ export function AverageGrade({
           <div className={styles.content}>
             <Tooltip.Arrow className={styles.arrow} />
             <p className={styles.title}>Average grade</p>
-            {average ? (
+            {passFailPercent !== null ? (
+              <p className={styles.description}>
+                Students received{" "}
+                <span style={{ color }}>{passFailPercent}% P</span> {tooltip}.
+              </p>
+            ) : average ? (
```
ARtheboss: I think the tooltip is not showing up on catalog. Not sure why

apps/frontend/src/app/Enrollment/index.tsx
```diff
@@ -306,24 +367,32 @@ export default function Enrollment() {
...
+                      color={output.color}
+                      subject={output.input.subject}
+                      courseNumber={output.input.courseNumber}
+                      enrollmentHistory={output.enrollmentHistory}
+                      hoveredDay={null}
+                      semester={output.input.semester}
+                      year={output.input.year}
+                    />
+                  </div>
+                ))
```
ARtheboss: Not sure what the issue is but HoverInfo seems to be broken right now

apps/frontend/src/components/ClassBrowser/List/index.tsx
```diff
@@ -18,27 +25,101 @@ interface ListProps {
...
   const { classes, loading } = useBrowser();
+  const [focusedIndex, setFocusedIndex] = useState<number>(0);
 
   const rootRef = useRef<HTMLDivElement>(null);
   const [searchParams] = useSearchParams();
 
+  // Keyboard navigation helpers
+  const { showFocusRing, showFocusRingTemporarily, hideFocusRing } =
+    useFocusRingTimer();
+  const isListFocused = useContainerFocus(rootRef);
```
ARtheboss: This gives me an error locally. It could potentially be null

apps/frontend/src/components/CourseBrowser/Filters/index.tsx
```diff
@@ -283,6 +283,39 @@ export default function Filters({
             </div>
           );
         })}
+        <p className={styles.label}>Level</p>
+        {Object.values(Level).map((level) => {
+          const active = currentLevels.includes(level as Level);
```
ARtheboss: This file will probably require more work to transfer over university/l&s requirements here too

PR #888, opened 2025-10-16
feat: Implement catalog cache refresh functionality

PineND:
- Added a new module for catalog cache refresh with scheduled execution.
- Integrated cache read/write logic in Apollo server configuration.
- Updated GraphQL resolver to support cache refresh option.

### DIFF
--- a/apps/backend/src/bootstrap/loaders/apollo.ts
+++ b/apps/backend/src/bootstrap/loaders/apollo.ts
@@ -83,6 +83,22 @@ export default async (redis: RedisClientType) => {
         defaultMaxAge: 24 * 60 * 60, // 24 hours
       }),
       responseCachePlugin({
+        shouldReadFromCache: async (ctx) => {
+          const variables = ctx.request.variables as Record<
+            string,
+            unknown
+          > | null;
+          if (variables?.refresh === true) return false;
+          return true;
+        },
+        shouldWriteToCache: async (ctx) => {
+          const variables = ctx.request.variables as Record<
+            string,
+            unknown
+          > | null;
+          if (variables?.refresh === true) return false;
+          return true;
+        },
         sessionId: async (req) =>
           req.request.http?.headers.get("sessionId") || null,
       }),
--- a/apps/backend/src/modules/catalog/cache-refresh.ts
+++ b/apps/backend/src/modules/catalog/cache-refresh.ts
@@ -1,0 +1,185 @@
+import { randomBytes } from "crypto";
+import os from "os";
+import type { RedisClientType } from "redis";
+
+import { TermModel } from "@repo/common";
+
+import {
+  buildCatalogCacheKey,
+  fetchCatalogData,
+  storeCatalogInCache,
+} from "./controller";
+
+const JOB_INTERVAL_MS = 5 * 60 * 1000;
+const LOCK_KEY = "locks:catalog-cache-refresh";
+const LOCK_TTL_SECONDS = 240;
+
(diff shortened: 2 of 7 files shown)


apps/backend/src/modules/catalog/cache-refresh.ts
```diff
@@ -0,0 +1,185 @@
...
+        error
+      )}`
+    );
+  } finally {
+    await releaseLock(redis);
+  }
+};
+
+export const registerCatalogCacheRefresh = (redis: RedisClientType): void => {
+  if (process.env.DISABLE_CATALOG_CACHE_REFRESH === "true") {
```
maxmwang: This env var should be added to `src/config.ts`. 

When would this variable be true?

apps/backend/src/bootstrap/loaders/apollo.ts
```diff
@@ -83,6 +83,22 @@ export default async (redis: RedisClientType) => {
...
+          return true;
+        },
+        shouldWriteToCache: async (ctx) => {
+          const variables = ctx.request.variables as Record<
+            string,
+            unknown
+          > | null;
+          if (variables?.refresh === true) return false;
+          return true;
+        },
```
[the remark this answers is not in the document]
ARtheboss: Seems like refresh is only used in one place and it's to set it to false anyway. You directly access the redis instead of using a query, so is this still necessary?

apps/backend/src/bootstrap/loaders/apollo.ts
```diff
@@ -83,6 +83,22 @@ export default async (redis: RedisClientType) => {
         defaultMaxAge: 24 * 60 * 60, // 24 hours
       }),
       responseCachePlugin({
+        shouldReadFromCache: async (ctx) => {
+          const variables = ctx.request.variables as Record<
+            string,
+            unknown
+          > | null;
```
maxmwang: This type should be `{ refresh: boolean } | null`. Below as well

apps/backend/src/modules/catalog/controller.ts
```diff
@@ -357,195 +585,47 @@ export const subjects: Record<string, Subject> = {
...
-
-    // TODO: Limit query because Fuse performance decreases linearly by
-    // n (field length) * m (pattern length) * l (maximum Levenshtein distance)
-    const filteredClasses = index
-      .search(query)
-      .map(({ refIndex }) => reducedClasses[refIndex]);
-
-    return filteredClasses;
+  if (cache) {
+    await storeCatalogInCache(cache, cacheKey, classes);
```
ARtheboss: probably no need to await

ARtheboss:
@PineND I'm not sure if this is the best way to do caching, as it is basically replicating what Apollo Server-Side caching already does. The refreshing data thing is a good idea but not sure if it can combine with Apollo caching.

apps/backend/src/modules/catalog/cache-refresh.ts
```diff
@@ -0,0 +1,185 @@
...
+  try {
+    const result = await redis.set(LOCK_KEY, INSTANCE_ID, {
+      NX: true,
+      EX: LOCK_TTL_SECONDS,
+    });
+
+    return result === "OK";
+  } catch (error) {
+    console.warn(
+      `[CatalogCache][${INSTANCE_ID}] Failed to acquire refresh lock: ${String(
```
maxmwang: Including `INSTANCE_ID` isn't really necessary since logs are separated by instance already.

PR #896, opened 2025-10-26
Scheduler Improvements

ARtheboss:
Removed caching and optimisticResponse stuff as it was causing a lot of strange errors. We should look into how to best add them back in a future PR.

New features

- Hover shows information for a section
- Show location for classes in default view
- Fixed reorder of sections
- Change color of classes


Chengming-Li (approved):
LGTM

PR #902, opened 2025-10-28
feat: P/NP display on frontend

vaclisinc:
## Problem Statement
Currently, if the course only exists P/NP, it won't show the percentage of pass.

## Result
By calculating the Pnp percentage in `apps/backend/src/modules/grade-distribution`, we can load the catalog page with XX% P but not impact the loading speed.
image


PineND:
Minor improvement, merging into catalog-pod for more tweaks.

PR #904, opened 2025-10-28
Feat/catalog grade pnp

PineND:
Add PNP logic to catalog display for classes that do not have grades otherwise.

CleanShot 2025-10-27 at 11  22 44@2x


PineND:
Added PNP logic on the backend, fix minor styling stuff for catalog too.

PR #905, opened 2025-10-28
Rename My Schedule to Scheduler, improved GradTrak and Enrollment sig…

PineND:
- Clicking on GrakTrak / Scheduler no longer automatically forces you to sign in. User get to see the Welcome screen with the option to sign in to proceed.
- Utilize new useUser hook to avoid redudant call to getUser (don't do much really).
- Rename My Schedules to Scheduler at the request of design.

CleanShot 2025-10-27 at 11  27 05@2x

### DIFF
--- a/apps/frontend/src/app/GradTrak/index.tsx
+++ b/apps/frontend/src/app/GradTrak/index.tsx
@@ -51,23 +51,7 @@ export default function GradTrakIndex() {
     }
   }, [user, userLoading, gradTraksLoading, hasGradTraks, gradTraks, navigate]);
 
-  if (userLoading) {
-    return (
-      <Boundary>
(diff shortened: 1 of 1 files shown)


apps/frontend/src/app/GradTrak/index.tsx
```diff
@@ -50,20 +51,26 @@ export default function GradTrakIndex() {
...
+    return (
+      <Boundary>
+        <LoadingIndicator size="lg" />
+      </Boundary>
+    );
   }
 
-  if (hasGradTraks) {
+  if (user && hasGradTraks) {
+    console.log("Loading user's GradTrak");
```
ARtheboss: remove

PR #922, opened 2025-10-31
enrollment notifications

PriyaDVenkatesan:
Backend and frontend for notifications feature on the user profile with user schema for 50%, 75%, 90% enrollment thresholds. Confirmation model should popup when a user unchecks all notifications and clicks anywhere else on the screen. Please check for hover states and dynamic color.


apps/backend/package.json
```diff
@@ -54,8 +56,10 @@
     "helmet": "^8.1.0",
     "keyv": "^5.5.3",
     "lodash": "^4.17.21",
-    "mongodb": "^6.20.0",
-    "mongoose": "^8.19.1",
+    "mongodb": "^6.18.0",
+    "mongoose": "^8.17.0",
```
ARtheboss: Is there a reason these are being downgraded?

maxmwang: This was probably an accident from merging in gql, we should just undo these changes

PR #968, opened 2025-11-14
Gql notifications backend

ritamzico:
All backend logic for the notifications feature. Add the api key to the .env file for it work. Frontend is on another branch.


apps/backend/src/modules/user/controller.ts
```diff
@@ -79,3 +79,33 @@ export const getBookmarkedClasses = async (
...
+
+    if (!_class) continue;
+
+    classes.push({
+      class: formatClass(_class),
+      thresholds: monitoredClass.thresholds,
+    });
+  }
+
+  return classes;
```
PineND: This can be improved a lot by by using a single query and doing join in the code. 

Example (double check, i just vibed this real quick):
```
// Helper to create a stable key for a class
const classKey = (c) =>
  `${c.year}-${c.semester}-${c.sessionId ?? "1"}-${c.subject}-${c.courseNumber}-${c.number}`;

// 1) Build list of query objects (or keys)
const classQueries = monitoredClasses.map((mc) => {
  const c = mc.class;
  return {
    year: c.year,
    semester: c.semester,
    sessionId: c.sessionId ?? "1",
    subject: c.subject,
    courseNumber: c.courseNumber,
    number: c.number,
  };
});

// 2) Single batched query
const classDocs = await ClassModel.find({ $or: classQueries }).lean();

// 3) Index by the same key
const classesByKey = new Map(
  classDocs.map((c) => [classKey(c), c])
);

// 4) Build final result, preserving thresholds
const result = [];

for (const monitoredClass of monitoredClasses) {
  const c = monitoredClass.class;
  const key = classKey(c);
  const doc = classesByKey.get(key);
  if (!doc) continue;

  result.push({
    class: formatClass(doc),
    thresholds: monitoredClass.thresholds,
  });
}

return result;
```

.gitignore
```diff
@@ -27,3 +27,9 @@ generated-types/
 # others
 *.cert
 *.key
+
+apps/frontend/tsconfig.node.tsbuildinfo
+apps/frontend/tsconfig.tsbuildinfo
+apps/frontend/vite.config.d.ts
+apps/frontend/vite.config.js
```
PineND: Suspicious, what produces these? why does our gitignore not handle it yet?

apps/datapuller/src/pullers/enrollment-checks.ts
```diff
@@ -0,0 +1,375 @@
...
+        }
+        <p style="margin: 5px 0; color: #666;">
+          <strong>Term:</strong> ${alert.semester} ${alert.year}
+        </p>
+      </div>
+    `
+    )
+    .join("");
+
+  const html = `
```
PineND: Abstract out to helper template file

infra/app/values.yaml
```diff
@@ -75,3 +75,6 @@ datapuller:
     enrollments:
       schedule: "0/15 * * * *"
       args: ["--puller=enrollments"]
+    enrollment-checks:
```
PineND: enrollment-checks is not needed - since `enrollments` is a cron job which already updates enrollments every 15 minutes, our notification system would plug into this - datapuller for enrollments can trigger the enrollment-check logic for us. Extend it instead.

packages/common/src/models/user.ts
```diff
@@ -75,6 +75,51 @@ export const userSchema = new Schema(
...
+      type: [
+        {
+          class: {
+            year: {
+              type: Number,
+              required: true,
+            },
+            semester: {
+              type: String,
+              enum: ["Spring", "Summer", "Fall", "Winter"],
```
PineND: Semester enum should already be defined somewhere

PineND (commented):
Brief review just looking through the code. Not comprehensive

apps/backend/src/modules/user/controller.ts
ritam2006: Used similar key construction to do a batched query instead querying for each class

.gitignore
Ryan6407: removed from the gitignore

apps/datapuller/src/pullers/enrollment-checks.ts
ritam2006: Moved to template file.

infra/app/values.yaml
ritam2006: Put all logic in enrollments job. Removed enrollment-checks.

PR #1052, opened 2026-02-07
discussion functionality

rcolato29:
Screenshot 2026-02-07 at 2 55 50 PM
Screenshot 2026-02-07 at 2 56 59 PM


apps/frontend/src/hooks/api/discussion/useAddComment.ts
```diff
@@ -0,0 +1,59 @@
+import { useMutation } from "@apollo/client/react";
+
+import { ADD_COMMENT, GET_COMMENTS } from "@/lib/api/discussion";
```
ARtheboss: Don't directly import. Use the generated `Document` types

PR #1083, opened 2026-03-01
finished adding bookmarks bar

marla-tumenjargal:
finished book marks bar within the figma onto the website


ARtheboss:
> _ No description provided. _

@marla-tumenjargal Please add a description so I know what this is

.env.template
ARtheboss: We've rotated the keys, but these changes are not correct. I cannot merge until you remove this.

ARtheboss (changes requested):

PR #1110, opened 2026-03-16
waitlist predictor on enrollment page

Jus973:
experimental feature, please test out and give feedback on UI + the overall feature. thinking about maybe pivoting to a less intrusive UI where it can show each course's difficulty to get into (sort by calculated lambda)

uses poisson-based regression from past 3 semesters for data. 

future inclusions might include:
only showing the calculator if the semester is the current semester
only showing the calculator if it is during course enrollment season

also, it might be possible to create this feature with less files(?) or neater(?), so i can clean the code up if you find a better way to do so.

### DIFF
--- a/apps/backend/src/modules/index.ts
+++ b/apps/backend/src/modules/index.ts
@@ -20,6 +20,7 @@ import Staff from "./staff";
 import TargetedMessage from "./targeted-message";
 import Term from "./term";
 import User from "./user";
+import Waitlist from "./waitlist";
 
 const modules = [
   Analytics,
@@ -42,6 +43,7 @@ const modules = [
   RouteRedirect,
   Pod,
   TargetedMessage,
+  Waitlist,
 ];
 
 export const resolvers = merge(modules.map((module) => module.resolver));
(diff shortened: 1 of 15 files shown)


vaclisinc (commented):
maybe we need to disable summer session:
Image

vaclisinc:
> @vaclisinc Because summer session enrollment is historically much different, should I completely remove the feature for it, or separate my data to only past summer sessions? although most classes wouldn't have much data.

I think it's fair we just support spring and fall semester. The summer sessions data probably is lesser and not stable (not sure but I guess every summer sessions may have rly different courses based on professors?)

vaclisinc (commented):
And also another UX aspect questions, once people add more than one class to the enrollment page, then how do the system know whether the user is inputing which course?

I think two solutions:
1. one is showing the course name and courseID on the top - quick solution for now?

2. maybe changing the arrangement (since we will probably make this into the waitlist volatility meter you mentioned before, and I lowk think this is maybe better than a number of probability at this moment - however, the volatility meter will be even bigger height and rn it's already not fitting in the 100% screen size) - maybe discuss with design team, I'll tag them later

PR #1126, opened 2026-04-03
Leo/relevance search

bfliao:
**Problem**:

Current catalog search returns results ranked purely by fuzzy match score. Users who regularly search for the same courses (e.g. their major requirements) have to scroll to the same results every session when they type the same search term. There's no personalization such that when users click on some classes in the past, these classes will be prioritized in the returned query.

**Scope**
Goals:

- Retention of click history (how many times an user visits a particular class )
- Re-rank search results after fuzzy search using history (local or server-side)
- Work for anonymous users (no login required)
- 

Non-Goals:

- Modification on fuzzy search - breach of abstraction
- Collaborative filtering ("people who took X also took Y")


bfliao:
@PineND merge?

PineND:
Implementation blocker is that cache control for pagination is now private, so everytime someone fetch the catalog via search we create a unique cached response. Hit rate basically go to 0 and our system will get a lot more load. We would have to seperate the query into 2 layer, a generic layer that have higher hit rate, then reranking after. Very complex and im not sure if its worth it

PR #1132, opened 2026-04-13
Ritam/unified tracking

ritamzico:
## Problem

  Two problems with old tracking:

  1. **Fragmented models.** Click events, banner views, and
  class views each had their own model (`ClickEvent`,
  `BannerViewCount`, `ClassViewCount`) with incompatible
  schemas. Adding a new event type meant a new model, new
  resolver, new migration. No shared query surface across
  event types.

  2. **No frontend batching.** Every tracking call fired its
  own network request immediately. High-traffic pages (catalog
   search, class view) hammered the backend with individual
  mutations.

  ## Solution

  Single `TrackingEvent` model with open string enums
  (`eventType`, `targetType`) and a flexible `metadata: JSON`
  field. Any team can add new event types without schema
  changes.

  Frontend `useTracking` hook queues events client-side and
  flushes in one `trackEvents` mutation — either every 5
  seconds or when 50 events accumulate, whichever comes first.
   Server enriches each event with `sessionId`, `ipHash`,
  `userAgent`, and `referrer` so clients send only what they
  know.

  ## What changed

  **Backend**
  - New `TrackingEvent` Mongoose model with indexes on
  `[eventType, targetType, timestamp]`, `[targetId,
  timestamp]`, `[sessionId, timestamp]`, and `[userId,
  timestamp]`
  - `trackEvents` GraphQL mutation — accepts up to 50 events
  per call, buffers to Redis, flushed to MongoDB every 5
  minutes
  - `trackingEventsTimeSeries` staff-only query for analytics
  - Dual-emit on all server-side redirect routes
  (`/banner/click/:id`, `/go/*path`, `/message/click/:id`) —
  old counters preserved, new model always written

  **Frontend**
  - `useTracking` hook with client-side batching
  (`trackClick`, `trackView`, `trackDismiss`, `trackSearch`,
  `trackSearchClick`)
  - Search tracking added to `CourseSearch` — debounced 1s,
  captures query + result count + which result was clicked
  - Banner view/dismiss and class view now emit to
  `TrackingEvent` alongside existing tracking
  - Staff analytics (`OutreachAnalytics`) reads from
  `TrackingEvent` via `trackingEventsTimeSeries`

  **Migration**
  - `apps/backend/src/scripts/migrate-click-events.ts`
  backfills all historical `ClickEvent` records into
  `TrackingEvent`. Safe to re-run (ignores duplicate key
  errors)

  ## What's preserved

  Old `clickCount`/`viewCount`/`dismissCount` counters on
  banners, redirects, and targeted messages are untouched. Old
   `ClickEvent` writes still happen when `clickEventLogging`
  is enabled on a resource. No data is deleted.


ritamzico:
I will create another PR after getting rid of the ClickEvent model and other such tracking infrastructure. The migration needs to be run on production first before doing that, so that steps needs to be separate.

apps/frontend/src/hooks/api/tracking/useTracking.ts
```diff
@@ -0,0 +1,118 @@
...
+import { useMutation } from "@apollo/client/react";
+
+import {
+  TrackEventsDocument,
+  TrackEventsMutation,
+  TrackEventsMutationVariables,
+  TrackingEventInput,
+} from "@/lib/generated/graphql";
+
+const BATCH_INTERVAL_MS = 5_000;
```
PineND: 5s interval is too aggressive, lets do 30s interval + pagehide beacon flush. 50 events cap is good to go

apps/backend/src/modules/tracking/controller.ts
```diff
@@ -0,0 +1,160 @@
...
+      try {
+        await TrackingEventModel.insertMany(documents, { ordered: false });
+        flushed += documents.length;
+      } catch (error) {
+        console.error("[TrackingEvents Flush] insertMany error:", error);
+        errors++;
+      }
+    }
+
+    await redis.lTrim(REDIS_BUFFER_KEY, raw.length, -1);
```
PineND: The lTrim runs always, if insertMany fails, up to 500 events are dropped from Redis and never written to Mongo. Either skip the trim on error, or inspect insertedIds and only trim the ones that actually landed.

apps/backend/src/modules/tracking/jobs/flush-tracking-events.ts
```diff
@@ -0,0 +1,31 @@
+import type { RedisClientType } from "redis";
+
+import { flushTrackingEvents } from "../controller";
+
+const FLUSH_INTERVAL_MS = 5 * 60 * 1000; // 5 minutes
+const STARTUP_DELAY_MS = 90 * 1000; // 1.5 minutes (staggered from other jobs)
+let isRunning = false;
```
PineND: isRunning here is per process. Lets move it to be redis lock. Redis is shared between different replicas, this variable is within each process. So race condition can happe here but not with redis

apps/backend/src/modules/tracking/jobs/flush-tracking-events.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const startTrackingEventsFlushJob = (redis: RedisClientType) => {
+  const runFlush = async () => {
+    if (isRunning) return;
+
+    isRunning = true;
+
+    try {
+      const result = await flushTrackingEvents(redis);
```
PineND, line 19: Lets also cap redis buffer key, if the flush job here is slow/breaks, or some how we cant write to DB, we get into a bad push state that appends forever.

apps/frontend/src/components/RootWrapper/index.tsx
```diff
@@ -44,9 +46,10 @@ export default function RootWrapper() {
       lastIncrementedPath = currentPath;
       lastIncrementTime = now;
       incrementClick(matchingRedirect.id);
+      trackClick("redirect", matchingRedirect.id, { fromPath: currentPath });
```
PineND, line 50: Apparently to claude: trackClick only enqueues. The nav happens immediately; the 5s timer never fires; the unmount flush uses Apollo mutate which the browser cancels mid-flight on full navigation. This event is effectively always lost. Fix: navigator.sendBeacon path for flush, wired into pagehide and also used synchronously before known navigations. (Same pattern would fix the earlier "is 5s too aggressive" question.

i didnt know this either

apps/frontend/src/hooks/api/tracking/useTracking.ts
```diff
@@ -0,0 +1,118 @@
...
+const MAX_BATCH_SIZE = 50;
+
+/**
+ * Unified tracking hook. Batches events client-side and flushes them
+ * via a single `trackEvents` GraphQL mutation every 5 seconds or when
+ * the batch reaches 50 events.
+ */
+export const useTracking = () => {
```
PineND: UseTracking() is called in Banner, Class, CourseSearch, RootWrapper. Each instance has its own queueRef and its own 5s timer, so a user interacting with several areas generates several small mutations instead of one batched call. Make it a module-level singleton (or a React context) so all callers share one queue.

apps/backend/src/modules/tracking/resolver.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (!staffMember) {
+    throw new GraphQLError("Only staff members can perform this action", {
+      extensions: { code: "FORBIDDEN" },
+    });
+  }
+
+  return staffMember;
+};
+
+const resolvers = {
```
PineND, line 42: Should explore how to prevent spams. Per IP/session rate limit, reject if the metadata is too large, key sanitization. Doubt anyone would majorly attack us but lets keep it safe

PineND (dismissed):

LehuyH:
LGTM!

ritamzico:
@PineND lmk if it looks good now

NathanDai5287 (dismissed):

NathanDai5287 (approved):
lgtm

PR #1138, opened 2026-04-23
Valeri/Miscellaneous-Block

Valeri-kula:
This is to create the miscellaneous bank. It will appear at the bottom of the page when you either have a class in drag state or you hover over the block with your mouse. When you move classes into it - it removes it from the total units and when you pull classes out of it into the semester blocks it would add the units to the total.


ARtheboss:
@Valeri-kula docker-compose.yml has some unrelated changes related to your .env stuff. Please remove that

PR #1141, opened 2026-05-01
[GradTrak] Optimal course-to-requirement matching with double major overlap enforcement

ethanransing:
# **Summary**
Introduces optimal course-to-requirement matching for GradTrak, which replaces greedy assignment with algorithms that optimally assign courses to requirements within a program (e.g. L&S's 7 Breadths), and also handle Berkeley's upper-division overlap policy for double majors. 

# **Changes**
1. Bipartite matching for breadth/count requirements. Replaces greedy matching 
2. Unit-threshold buckets (ex. >= X units of UD classes): Course assignment uses backtracking w/ pruning to optimally fulfill buckets
3. Joint matching across double majors (see [L&S](https://lsadvising.berkeley.edu/majors-minors/double-majors-simultaneous-degrees), [CDSS](https://cdss.berkeley.edu/academics/policies/majors-and-minors), [CoE](https://engineering.berkeley.edu/academics/majors-and-minors/major-options/) policies):  Backtracking w/ pruning to assign upper division classes to major A, major B, both (if overlap cap isn't exceeded), or none 
4. Testing: Unit tests in packages/BtLL/src/test.ts covers 
5. UX change: hide DS domain emphasis subrequirements by default

# **Limitations**
1. Joint majors aren't handled. These don't have the same overlap restrictions as double majors
2. This CoE double major restriction isn't encoded: "Each major must include at least 5 upper division technical courses, of at least 3 units each, that do not overlap with the other major." 
3. 3+ majors aren't supported


ethanransing:
splitting this into a bunch of smaller prs

PR #1146, opened 2026-05-19
Fix/reviews

Henrp:
Fixed scroll on grades and professor option. Added pass/no pass option to grades option and removed F- and F+.


ARtheboss:
What about removing overall rating?

Henrp:
will remove once i figure out how the card should look cause it looks squished if i remove the number

package.json
ethanransing, line 1: these changes seem unintentional?

apps/frontend/src/components/Class/Ratings/RatingDetail/RatingDetail.module.scss
```diff
@@ -23,7 +23,7 @@
 
         .title {
           margin: 0;
-          color: var(--paragraph-color);
+          color: rgb(64, 64, 64);
```
ethanransing: is this robust to light/dark mode? screenshots here would be nice

packages/shared/ratingsConfig.ts
```diff
@@ -7,7 +7,7 @@ export const USER_MAX_ALL_RATINGS = 50;
 export const USER_MAX_SEMESTER_RATINGS = 8;
 
 // ratings required from a user before unlocking ratings tab
-export const USER_REQUIRED_RATINGS_TO_UNLOCK = 0;
+export const USER_REQUIRED_RATINGS_TO_UNLOCK = 1;
```
ethanransing, line 10: why do we do this now? just curious

apps/backend/src/modules/rating/controller.ts
```diff
@@ -524,14 +524,39 @@ export const getUserRatings = async (context: RequestContext) => {
     }
   });
 
+  const classNumbers = formattedUserRatings.classes.map((c) => c.classNumber);
+  const sections = await SectionModel.find({
+    classNumber: { $in: classNumbers },
+  }).select("semester year classNumber meetings");
+
+  const instructorNamesByClassKey = new Map<string, string>();
```
ethanransing: this logic is mirrored in getClassRatings.ts and have different deduplication strategies (the map), should consolidate

packages/theme/src/components/Select/index.tsx
```diff
@@ -610,6 +610,10 @@ export function Select<T>({
                   style={
                     maxListHeight ? { maxHeight: maxListHeight } : undefined
                   }
+                  onWheelCapture={(e) => {
+                    e.stopPropagation();
+                    e.currentTarget.scrollTop += e.deltaY;
```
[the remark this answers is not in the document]
ethanransing: codex seems correct here (at least my claude chromedevtools mcp agrees). can we just get rid of the manual scrolling?

packages/shared/ratingsConfig.ts
Henrp: image

package.json
```diff
@@ -35,6 +35,12 @@
   },
   "name": "berkeleytime",
   "dependencies": {
+    "@opentelemetry/api": "^1.9.1",
```
ethanransing: i think this is a local env artifact

apps/backend/src/bootstrap/index.ts
```diff
@@ -1,4 +1,4 @@
-import express from "express";
```
ethanransing, line 1: this seems unintentional

apps/backend/src/modules/rating/controller.ts
```diff
@@ -524,14 +566,23 @@ export const getUserRatings = async (context: RequestContext) => {
     }
   });
 
+  const classNumbers = formattedUserRatings.classes.map((c) => c.classNumber);
+  const sections = await SectionModel.find({
```
ethanransing, line 570: this is an unindexed scan through the db. we should scope the query or just copy what getClassRatings does to avoid a linear time scan

apps/backend/src/bootstrap/index.ts
Henrp: backend typecheck fails when i keep this cause its saying its not used so i remove

ethanransing (dismissed):
i think you can fix the package-lock.json diff by re running npm install

ethanransing (approved):

PR #1149, opened 2026-05-31
Add backend runtime metrics + bound heap/probes

johngerving:
### Telemetry

- instrumentation.ts — add RuntimeNodeInstrumentation to the OTel SDK, exporting event-loop delay/utilization, GC duration, and V8 heap metrics.
- metrics.ts — add self-contained gauges built only on Node built-ins (perf_hooks + process.memoryUsage), so they report even before the new dependency is installed: process.eventloop.lag.max / .p99 (ms) and process.memory.heap_used / .rss (bytes). The two event-loop gauges share one batched callback so both read the same snapshot before the histogram is reset each export window.
- package.json — declare @opentelemetry/instrumentation-runtime-node@^0.13.0 (aligned with the existing 0.56-era OTel stack; peer dep is only api ^1.3.0, which is satisfied).

### Infra (`infra/app/templates/backend.yaml`)

- NODE_OPTIONS=--max-old-space-size=1024 so V8 GCs at a predictable bound rather than growing toward node RAM.
- Memory request: 768Mi / limit: 1536Mi (heap cap plus headroom for zlib/buffers/native), CPU request: 500m, and no CPU limit (CFS throttling would stall the single event-loop thread).
- A livenessProbe deliberately more lenient than readiness (~150s before restart): transient lag de-registers a pod from the Service, but only a genuine multi-minute wedge restarts it.

### DIFF
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -44,6 +44,7 @@
     "@opentelemetry/sdk-logs": "^0.56.0",
     "@opentelemetry/sdk-metrics": "^1.29.0",
     "@opentelemetry/sdk-node": "^0.56.0",
+    "@opentelemetry/instrumentation-runtime-node": "^0.13.0",
     "@apollo/server-plugin-response-cache": "^5.0.0",
     "@apollo/utils.keyvadapter": "^4.0.1",
     "@as-integrations/express5": "^1.1.2",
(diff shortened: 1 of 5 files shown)


MatthewZhu13 (approved):
Looks good to me

PR #1151, opened 2026-06-09
Fix "Open in Grades" in Catalog: pass courseId instead of courseNumber to fix mismatch

ethanransing:
## Issue:
image
image<br>

Another issue: When you open a class in grades and refresh, the name of the course is a blank string: 
image

### Root cause: 
#1130 changed the backend/frontend contract by adding courseId, but the "Open in Grades" button contract wasn't updated to match this. Instead courseNumber was passed in as courseId, which leads to no match. #1130 also changed the URL writer to not include course number, so on refresh the URL reader can't reconstruct the course number from the course ID in the url. 
### Fix:
1) Pass in courseId to the button
2) Add GetCourseNumberByIdDocument query


Henrp (approved):
looks good to me

PR #1153, opened 2026-07-03
fix: pull FUTURE courses and dedupe to latest version

ritamzico:
## Summary

The course datapuller dropped `FUTURE` (upcoming-term) courses, so classes pointing to
them never showed up in the catalog. Now keeps `ACTIVE` + `FUTURE`.

## Changes

- `filterCourse`: keep `ACTIVE` and `FUTURE`.
- `getCourses`: dedupe to newest `fromDate` per `{courseId, subject, number}`.


sashakmurray (approved):
lgtm

PR #1154, opened 2026-07-14
Set max memory limit = 140mb

sashakmurray:
the `save ""` setting disables RDB snapshotting so that container restarts don't result in re-loading the data from disk every time (which is what is currently causing the repeated crashes)


Henrp (approved):
good

PR #1156, opened 2026-08-02
chore: seal SMTP_PASSWORD for backend and datapuller

bfliao:
Updated helm chart

Adds kubesealed `SMTP_PASSWORD` to the backend and datapuller SealedSecrets so the credential is injected automatically on every helm upgrade


Henrp (approved):
lgtm

PR #1157, opened 2026-08-12
fix: make curated classes findable by their real course number

ritamzico:
Advisors reported that MUSIC 32, INTEGBI 130 and EALANG 121 could not be added in the Curated Classes tool. These turned out to be two unrelated bugs, plus a third found while verifying.

All three classes and their primary sections exist in `classes`/`sections` for Fall 2026. What breaks is the join to `courses`, which both catalog paths perform by `courseId`.

## 1. The tool identified classes by the course record

`CuratedClassEditor` took `subject`/`courseNumber` from the joined **course**, not the class. SIS renumbered `courseId 111172` from INTEGBI 32 to INTEGBI 130 effective 2026-08-19, carried only on the course's FUTURE version, so the class showed up as "INTEGBI 32" and was unfindable by its real number.

This is wider than renumbers. In Spring 2026 alone, **371 of 16,086 classes** have a class-level subject/number differing from the joined course record — mostly cross-listings. Running the editor's own search over that data:

```
"AFRICAM C143A"  before: (no results)          after: AFRICAM C143A #001
"AMERSTD C152"   before: (no results)          after: AMERSTD C152 #001
"THEATER C183A"  before: AFRICAM C143A #001    after: (no results)
```

It also affected saving: those values are written into the curated class, and the backend resolves them against `ClassModel`, so an entry saved as "INTEGBI 32" would never have resolved.

The list page had a matching bug — rows render `{subject} {courseNumber}` but the search index was built from `{subject} {number}`, the *section* number, so no course number ever matched ("EALANG C120" found nothing; "EALANG 001" found it).

## 2. The courses puller has not written anything in months

The puller fetches ACTIVE/FUTURE courses, then deleted every stored course whose `courseId` was absent from that fetch — 42,304 of 58,399 docs, including 10,336 courseIds still referenced by past-term classes. A ratio guard stood in front of it and returned early whenever the fetch was under 95% of `countDocuments()`. The collection holds one doc per `(courseId, subject, number)` while the fetch is deduped to current numbers, so the ratio sits at 16,254/58,399 = **0.28** and the guard aborted every run, silently, exit code 0. That is why no course created since has ever landed, including MUSIC 32 (courseId 168547) and EALANG 121 (168502).

Now it deletes only superseded numbers — docs whose `courseId` is in the fetch but whose `(courseId, subject, number)` is not — which clears leftovers like COLWRIT 6A once COLWRIT 22B lands, while keeping discontinued courses so past-term joins survive. With nothing destructive left to guard, the ratio check goes away. Failures now throw so the CronJob fails visibly instead of reporting success without writing. Page fetches also retry 3x with backoff before counting as errors, since a dropped page silently truncates the list.

## Verification

Ran against local Mongo:

```
Fetched 16,254 courses.
Deleted 111 superseded courses.
Completed updating database with 16,254 courses.
```

`courses` went 58,399 -> 58,558; INTEGBI 130, MUSIC 32, EALANG 121 and ARCH 101B all present, history intact. Also verified on an empty collection (guard skipped, 0 deleted, 16,254 upserted). Confirmed against SIS that all three classes' primary sections have `printInScheduleOfClasses: true`, so they clear the stricter filter the denormalized catalog applies.

### DIFF
--- a/apps/ag-frontend/src/app/CuratedClasses/index.tsx
+++ b/apps/ag-frontend/src/app/CuratedClasses/index.tsx
@@ -27,7 +27,7 @@ function Content({ curatedClasses }: ContentProps) {
 
   const fuzzySearch = useMemo(() => {
     const list = curatedClasses.map((c) => ({
-      name: `${c.subject} ${c.number}`,
+      name: `${c.subject} ${c.courseNumber}`,
       text: c.text,
       _id: c._id,
     }));
--- a/apps/ag-frontend/src/components/CuratedClassEditor/index.tsx
+++ b/apps/ag-frontend/src/components/CuratedClassEditor/index.tsx
@@ -83,7 +83,7 @@ export default function CuratedClassEditor({
     if (!classes?.catalog) return null;
 
     const list = classes.catalog.map((cls, index) => ({
-      name: `${cls.course.subject} ${cls.course.number}`,
+      name: `${cls.subject} ${cls.courseNumber}`,
       title: cls.title || cls.course.title || "",
       index,
     }));
@@ -108,8 +108,8 @@ export default function CuratedClassEditor({
     () =>
       classes?.catalog.find(
         (cls) =>
-          cls.course.subject === localValue.subject &&
-          cls.course.number === localValue.courseNumber &&
+          cls.subject === localValue.subject &&
+          cls.courseNumber === localValue.courseNumber &&
           cls.number === localValue.number &&
           cls.sessionId === localValue.sessionId
       ),
@@ -162,8 +162,8 @@ export default function CuratedClassEditor({
                 <Flex p="1" gap="3">
                   <Flex direction="column" flexGrow="1">
                     <Heading size="2">
-                      {selectedClass.course.subject}{" "}
-                      {selectedClass.course.number} #{selectedClass.number}
+                      {selectedClass.subject} {selectedClass.courseNumber} #
+                      {selectedClass.number}
                     </Heading>
                     <Text size="2" color="gray">
                       {selectedClass.title || selectedClass.course.title}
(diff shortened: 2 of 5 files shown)


NathanDai5287 (approved):

NathanDai5287 (approved):
## Frontend changes (ag-frontend)

### `GET_CATALOG` query change -- safe

`GET_CATALOG` is only consumed by `CuratedClassEditor`. Moving `subject`/`courseNumber` from the `course` fragment to the class level, and removing `course.subject`/`course.number`, will not break anything else. The backend `formatClass` already spreads these fields from the `IClassItem` document, so the resolver returns them at both levels. Since no other code depends on `course.subject`/`course.number` from this specific query, the removal is clean.

### `CuratedClasses/index.tsx` -- correct

The search index was built with `c.number` (section number), not `c.courseNumber`. Changing to `c.courseNumber` matches the `ICuratedClass` type and makes search-by-course-number work.

### `CuratedClassEditor/index.tsx` -- correct

All six references to `cls.course.subject`/`cls.course.number` are replaced with `cls.subject`/`cls.courseNumber`. The remaining `cls.course.title` access is fine since `title` stays in the `course` fragment.

---

## sis-api.ts -- retry logic

### Retry + backoff: correct

`fetchPage` retries up to 3 times with 500/1000/1500ms linear backoff. On the 3rd failure it re-throws, which the outer `.catch` in `fetchBatch` handles (logs, records in `failedPages`, returns `[]`). The flow is sound.

### 404-as-end-of-data: worth a second look

`isEndOfData` treats any 404 as "no more data" and returns `[]`. If the SIS API returns 404 for a valid page due to a transient issue (rather than genuinely meaning "past the last page"), this silently truncates results. The original code also swallowed all errors, so this is strictly better -- just noting the assumption.

### failedPages.push() thread safety: fine

Node.js is single-threaded. `Promise.all` resolves microtasks sequentially on the event loop, so concurrent `.push()` calls from different resolved promises cannot interleave. No race condition.

### Interaction between failedPages and early-stop heuristic

`processBatch` still uses `errorCount < queryBatchSize / 10` to decide whether to continue fetching. If a batch has many failures, pagination stops early AND `failedPages` triggers the final throw. Both mechanisms are consistent.

---

## courses.ts -- deletion logic

### Superseded-course identification: correct

The `courseKey` approach (`courseId|subject|number`) correctly identifies stored records that share a `courseId` with a fetched course but have a different subject/number combination. These are the renumbered courses that need cleanup.

### Deliberate scope reduction

The old code deleted any course whose `courseId` was absent from the fetch (`$nin`). The new code only deletes courses whose `courseId` IS present but whose subject/number changed. Courses that disappear entirely from SIS are now retained. This matches the stated intent (removing only superseded numbers), but if full cleanup of removed courses is ever needed, it would require a separate mechanism.

### Throwing on zero courses: good

Replacing the silent `return` with `throw` ensures the caller knows the pull failed rather than silently succeeding with no data.

---

No blocking issues. The changes are consistent and correctly scoped.

PR #1161, opened 2026-08-21
Fix catalog sort being ignored while a search query is active

davidmaking:
## Problem

I noticed that sorting by units/average grade/open seats did nothing whenever you had an active search query. Sorting in the catalog only worked when the search box was empty. As soon as any search text was entered, the Sort By dropdown kept updating and the URL would change to reflect whatever was put in, but it didn't actually do anything. At least for me, that got rid of most of the utility of the search function.

## Root cause

- Looking back at the commit history, I noticed that it seemed to be tied to an old implementation of search (PR #1101 built the search path on MongoDB Atlas $search) that didn't support search + sort.
- PR #1102 changed things to switch out the Atlas $search for a Fuse.js in-memory index, but the workaround wasn't changed, which meant search + sort stayed inactive even when it didn't seem necessary anymore.
- useCatalogQuery made sortBy/sortOrder undefined whenever there was a search query, and either way, the new(ish) Fuse.js search path (getCatalogWithSearch) never applied a sort to begin with.

## Fix

- useCatalogQuery will now always send sort state
- Sort happens between filtering and pagination so that following pages will be correct
- Tie-break by subject/courseNumber, which is the same tiebreaking methodology that regular non-search buildSort already uses
- When sorting by relevance, it doesn't actually sort by relevance, which doesn't exist, it just keeps Fuse's ordering by how well the output matches the query

## Testing

- I type-checked and linted the code
- I ran the page & dev stack (Docker: Mongo/Redis/backend/frontend) locally on my laptop with synthetic data since I didn't have access to actually query for real data, and it worked as expected
- I didn't see any test coverage for this code, so I didn't add any

### DIFF
--- a/apps/backend/src/modules/catalog/controller.ts
+++ b/apps/backend/src/modules/catalog/controller.ts
@@ -121,6 +121,8 @@ export const getCatalogSearch = async (params: CatalogQueryParams) => {
       semester,
       searchTerm: search.trim(),
       filters,
+      sortBy,
+      sortOrder,
       limit: effectivePageSize,
       skip,
     });
@@ -202,6 +204,8 @@ const getCatalogWithSearch = async ({
   semester,
   searchTerm,
   filters,
+  sortBy,
+  sortOrder,
   limit,
   skip,
 }: {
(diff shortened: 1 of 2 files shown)


davidmaking:
@sashakmurray Hi, I opened this PR to hopefully fix a little bug(?) in the catalog and contribute something to a tool I use a lot. Obviously I don't have write access to trigger the dev deploy that's described as the next step in contributing.md, so hopefully you can take a look.

ritamzico (approved):
Looks good to me; thanks for catching this issue and taking the initiative to submit a PR!

I'll merge this into main and deploy the change shortly.

If you're interested in getting involved with the Berkeleytime team, we'd love to hear from you. You can apply here: https://docs.google.com/forms/d/e/1FAIpQLSd8d8pqh9E_NtiX82GxnBUQpmIqXQcwIs8YuZhA2asPF2KU2A/viewform

Thanks again for the contribution!

PR #1162, opened 2026-08-21
Renamed Courses: Find a Course by Its Old Name

NathanDai5287:
## What changed

The university sometimes renames a course. For example, "EECS 16A" became "ELENG 66" starting Fall 2026. Students who knew the old name had no way to find the course: typing "EECS 16A" into a course search brought up unrelated courses, and the renamed class looked like it had vanished.

Now, searching the old name finds the renamed course. This works in the course picker on the Enrollment and Grades pages and in the Catalog search.

| | Before | After |
|---|---|---|
| Course picker (Enrollment/Grades), searching the old name "EECS 16A" | Before: only an unrelated course comes up | After: ELENG 66 is the top suggestion |
| Catalog search (Fall 2026), searching "EECS 16A" | Before: ELENG 66 is missing from results | After: ELENG 66 is the top result |

## Test plan

- [ ] Search "EECS 16A" in the Enrollment or Grades course picker → ELENG 66 appears and can be selected
- [ ] Search "EECS 16A" in the Catalog with the Fall 2026 term selected → ELENG 66 is the top result
- [ ] Search courses by their current names → results unchanged

### DIFF
--- a/apps/backend/src/modules/course/controller.ts
+++ b/apps/backend/src/modules/course/controller.ts
@@ -40,17 +40,38 @@ const buildFormerNamesByCourseId = async () => {
   );
 };
 
-let formerNamesByCourseIdCache: Promise<Map<string, string[]>> | null = null;
+const FORMER_NAMES_CACHE_TTL_MS = 5 * 60 * 1000;
+
+interface FormerNamesCacheEntry {
+  promise: Promise<Map<string, string[]>>;
+  expiresAt: number;
+}
+
+let formerNamesByCourseIdCache: FormerNamesCacheEntry | null = null;
 
 export const getFormerNamesByCourseId = () => {
-  formerNamesByCourseIdCache ??= buildFormerNamesByCourseId().catch(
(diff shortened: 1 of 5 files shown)


apps/backend/src/modules/course/controller.ts
```diff
@@ -10,6 +10,49 @@ import { formatClass } from "../class/formatter";
...
+
+  return new Map(
+    classNames.map(({ _id, names }) => [
+      _id,
+      names.filter((name) => !currentCourseNames.has(name)),
+    ])
+  );
+};
+
+let formerNamesByCourseIdCache: Promise<Map<string, string[]>> | null = null;
```
johngerving: Should this be invalidated at some point?

NathanDai5287: good catch. i added a 5 min TTL, which matches the cache TTL for the rest of course search

johngerving (approved):
LGTM

PR #1170, opened 2026-08-27
Hide recruitment banner on Apply page

NathanDai5287:
## Summary
- keep the standard navigation and footer on \/apply\`n- suppress the global outreach banner on that page so it does not advertise the page from within itself

## Verification
- \
pm run type-check --workspace=apps/frontend\`n- headless Chromium check confirmed Apply content and navigation remain visible while the recruitment banner is absent


NathanDai5287:
Closing in favor of the existing feat/apply-page branch, per request.

PR #1173, opened 2026-08-30
Activate BtLL test suite and wrap BtLL in try/catch blocks

ethanransing:
The BtLL test suite couldn't fail. The runner printed pass/fail but never set an exit code, so a broken test still exited 0 — wiring it into CI as-is would have given us a permanently green check. This makes it exit non-zero on failure, quiets the debug output so a failure is actually readable, fixes two tests that were asserting the wrong thing, and adds a test task to turbo so it runs.

Chasing that turned up a related bug: if any single requirement program throws while being evaluated, the whole GradTrak requirements panel goes blank, with nothing logged. A bad program is now logged and skipped, and the remaining requirements still render.

### DIFF
--- a/apps/frontend/src/app/GradTrak/BtLLInterface/index.tsx
+++ b/apps/frontend/src/app/GradTrak/BtLLInterface/index.tsx
@@ -623,9 +623,15 @@ export default function BtLLGradTrakInterface({
         ]);
 
         const config = { debug: false, fetchCourse };
-        const evaluated = init(req.code, commonVars, config) as
-          | RequirementResult[]
-          | null;
+        let evaluated: RequirementResult[] | null;
+        try {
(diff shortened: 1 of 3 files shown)


Henrp (approved):
I tested and it threw an error as expected. Also, I found that the UC Requirement disappears when there is less than 3 semester in GradTrak. Was this intended?

ethanransing:
> Also, I found that the UC Requirement disappears when there is less than 3 semester in GradTrak. Was this intended?

this is to avoid throwing, iirc something to do with the senior residence requirement but ill double check before merging

ethanransing:
> I tested and it threw an error as expected. Also, I found that the UC Requirement disappears when there is less than 3 semester in GradTrak. Was this intended?

addressing in https://github.com/asuc-octo/berkeleytime/pull/1214

PR #1174, opened 2026-08-31
Collapse GradTrak domain emphases by default + fix vertical spacing 

ethanransing:
Students only complete one domain emphasis, so expanding every one by default buries the rest of the requirements. This collapses them, and fixes the spacing between the rows — using the spacing scale the app already loads rather than more hardcoded values.

Also fixes a typo in the side panel disclaimer.

**Before**
before

**After**
after

**With one emphasis expanded**
after expanded


Henrp (approved):
good

PR #1175, opened 2026-08-31
Build Course from one place, add sum_units

ethanransing:
The BtLL Course constructor had different return shapes for the two forms it accepts (`{"CS 61A"}`, `{"CS", "61A"}`). The two-argument shape is missing a field the type declares as required, causing the field to throw. 

Also adds `sum_units`, which the requirement programs currently hand-roll as a reduce in a couple dozen places, and exports `coursesEqual`. Both are needed by the course-assignment work this is split out from.


Henrp (approved):
lgtm

PR #1176, opened 2026-08-31
Rudy/explore

rcolato29:
This PR adds the Explore page to Berkeleytime, with rows for popular, curated, subject-based, and personalized picks. Recommendations are served via the semantic-search service and are ranked + de-duped across rows. Evaluation and testing are included in order to ensure both sufficient recommendation quality and sound ranker logic. Hand-picked recommendations are pulled from the Curated page, with the latest term prioritized. Popular courses are built from Berkeleytime's course ratings and use Bayesian shrinkage to ensure that popularity is statistically grounded. The main rows are built from ENN search with recency weighting. All features work for anonymous users (since course history is client-side), so no login is necessary.


vaclisinc:
what a crazily huge PR lmfao

PR #1181, opened 2026-09-03
address ISP

sashakmurray:
- docker binds to port 127.0.0.1 instead of 0.0.0.0
- redis and mongo now require generated credentials (run `node scripts/initialize-local-env.mjs` to generate)
- CI enforces these security requirements

In order to enable auth for Mongo, we need to remove the existing volume. The bootstrap-local script now includes generating the redis/mongo credentials.
```
docker compose down --volumes
docker compose pull
bash apps/docs/src/getting-started/bootstrap-local.sh
```


Henrp (approved):
lgtm

PR #1188, opened 2026-09-05
fixed type mismatch for ratings aggregator

Henrp:
Whenever a user submitted a rating, userRatingsAggregator didn't compare context.user._id against the right data type. This PR fixes that by converting context.user._id to a string before the comparison so it actually works.

### DIFF
--- a/apps/backend/src/modules/rating/helper/aggregator.ts
+++ b/apps/backend/src/modules/rating/helper/aggregator.ts
@@ -215,7 +215,7 @@ export const termsRatingsAggregator = async (
 export const userRatingsAggregator = async (context: RequestContext) => {
   if (!context.user?._id) throw new Error("Unauthorized");
   return await RatingModel.aggregate([
-    { $match: { createdBy: context.user._id } },
+    { $match: { createdBy: context.user._id.toString() } },
     {
       $group: {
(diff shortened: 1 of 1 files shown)


ethanransing (commented):
why do we need this? https://github.com/asuc-octo/berkeleytime/blob/721d41c7724eb660bad5f91d0d790292d9f7b907/apps/backend/src/types/request-context.ts

looks like user id is a string already?

Henrp:
thats the declared type; in line 177-192 of file apps/backend/src/bootstrap/loaders/passport.ts, what i think it does is it never converts to string and this is the part that is actually assigns the user id. I don't really know this too well so ill also include claude explanation:

> context.user._id is typed as string but is actually a live Mongo ObjectId (it's refetched per-request in deserializeUser via .findOneAndUpdate().lean()), while RatingModel.createdBy is a String field. Regular find/findOne calls auto-cast that mismatch away, but aggregate()'s $match doesn't, so it silently matched nothing until we added .toString() — that's the root cause of the count mismatch.

ethanransing-aven (approved):
we should figure out whats actually goijg on lol

ethanransing (approved):
wrong account 🥶

PR #1199, opened 2026-09-07
Fix term refreshes resetting catalog availability

johngerving:
### Summary

Prevents the terms datapuller from temporarily hiding active semesters from Catalog.

### Problem

Term refreshes deleted and reinserted nearby terms with hasCatalogData: false. Catalog only displays terms where this flag is true, so semesters disappeared until the classes datapuller recalculated it 45 minutes later.

### Changes
- Replace term deletion and reinsertion with upserts.
- Preserve hasCatalogData for existing terms.
- Initialize new terms with hasCatalogData: false until class data is available.


MatthewZhu13 (approved):
LGTM

PR #1200, opened 2026-09-07
Fix semantic search to index only terms with catalog data

johngerving:
## Summary

Fixes semantic search startup selecting empty future terms by filtering term discovery to terms with catalog data.

## Changes

- Query terms with `withCatalogData: true`
- Regenerate persisted GraphQL operations
- Add regression coverage ensuring semantic search and catalog use the same populated terms


sashakmurray (approved):
looks good!

PR #1201, opened 2026-09-09
P/np Percentile Update

MatthewQ1:
P/NP grades currently count as being under F's in percentile calculation. This change makes P/NP grades ignored for percentile calculation.


MatthewQ1:
Note that dev environment has been broken since June 15.

NathanDai5287:
@copilot does this have side effects anywhere other than the visualization in the grades graph?

NathanDai5287 (approved):

PR #1209, opened 2026-09-13
Fix semantic search index rebuild authorization and eviction safety

johngerving:
Addresses [BER-5](https://linear.app/berkeleytime/issue/BER-5/anyone-can-rebuild-or-delete-course-search-indexes-without-logging-in)

### Summary
- Restrict semantic search index refreshes to authenticated staff
- Add a five-minute Redis-backed refresh rate limit
- Prevent searches for missing indexes from triggering background builds
- Preserve production indexes by evicting based on distinct academic terms
- Add tests for authorization, throttling, rebuild prevention, and eviction

### DIFF
--- a/apps/backend/src/modules/semantic-search/security.test.ts
+++ b/apps/backend/src/modules/semantic-search/security.test.ts
@@ -1,0 +1,123 @@
+import assert from "node:assert/strict";
+import { test } from "node:test";
+
+import { createRefreshRateLimit, requireStaffRefreshAccess } from "./security";
+
+type ResponseSnapshot = {
+  body?: unknown;
+  headers: Record<string, string>;
+  status?: number;
+};
+
+function responseStub(snapshot: ResponseSnapshot) {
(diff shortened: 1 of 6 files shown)


apps/backend/src/modules/semantic-search/security.test.ts
sashakmurray, line 1: I'm unsure that all of these tests are necessary. Same with the python tests. However that decision is up to you.

My other question is, where / how are these tests being run? It seems like they're not included in CI, so would we rely on developers to run them locally? maybe I'm missing something.

sashakmurray (commented):
change seems good, see comment

apps/backend/src/modules/semantic-search/security.test.ts
johngerving: You're right, they were mostly for me to run locally. I'll get rid of them.

sashakmurray (approved):

PR #1214, opened 2026-09-13
Fix UC Requirements disappearing on plans with fewer than 3 Fall/Spring terms

ethanransing:
Before #1173 this threw and made all course requirements vanish. After #1173, only the UC Requirements group vanished. The root cause is an indexing error, since we get the columns with `get_element(only_spring_fall, length - 3)`, which breaks when the course plan only has 1-2 semesters. However this is in an if/else block that shouldn't be evaluated with 1-2 course semesters.

The root cause is that BtLL evaluates both blocks in if/else statements, rather than lazily like every other language. We need to evaluate only the block that will run to make get_element for this application safe. 

Added unit tests that verify that we don't evaluate something that would throw in if/else and elseif blocks. Also added a BtLL test that runs `UC_REQ_BTLL` against a two-semester plan; fails on `main`, passes here.

The frontend evaluates `PlanRequirement.code` from Mongo, so the UC doc needs to be re-seeded (`npm run seed:requirements` in `packages/BtLL`) for this to take effect.


Henrp (approved):
tested and works now. 

I noticed when i was trying to understand this bottom up, the planTerm from MongoDB is not sorted in requirements panel. A problem this causes is senior residence check count. For example if you have Fall 23, Spring 24, Fall 24, Spring 25, Summer 21 <- this gets counted towards the seinor residence. its sorted in the gradtrak ui cause the semester grid copies the planTerm and adds a .sort()

ethanransing:
> I noticed when i was trying to understand this bottom up, the planTerm from MongoDB is not sorted in requirements panel. A problem this causes is senior residence check count. For example if you have Fall 23, Spring 24, Fall 24, Spring 25, Summer 21 <- this gets counted towards the seinor residence. its sorted in the gradtrak ui cause the semester grid copies the planTerm and adds a .sort()

is there any reason to not move this sort to the backend? this seems dumb

Henrp:
yeah we can move it to make it cleaner.

PR #1216, opened 2026-09-14
Remove all pre-push hooks except format and lint.

sashakmurray:
These hooks were very flaky and most developers skipped them. They also require building the entire application even if the change was trivial. Instead, we run these checks in CI.


bfliao:
Tried this branch out — pushes went from multi-minute to ~6s, big improvement. One thing I hit while testing that I think belongs in this PR: the surviving `format` hook will cause git push to fail even after Prettier runs. It predates this PR, but this is the natural place to fix it.

### Why `--write` fails the push although it fixed your file

`format` runs `npx prettier --write`, which edits files. pre-commit fails any hook that modifies files. So if your code isn't already Prettier-clean:

```
format...................................................................Failed
- hook id: format
- files were modified by this hook
```

### And retrying the push never works

This failed push left Prettier's fix in your working tree as an uncommitted edit. On the next push, pre-commit stashes it, reruns Prettier, and can't reapply the stash cleanly:

```
[WARNING] Stashed changes conflicted with hook auto-fixes... Rolling back fixes...
```

So push #2 fails, and #3, and #4 — identically, indefinitely. I verified the file's md5 is frozen across repeated attempts. The normal instinct after a failed push is to push again, and that instinct can never succeed here. You have to notice the ` M` in `git status` and commit Prettier's fix instead. Nothing in the output says so.

Repro on this branch — no unusual setup, just commit and push:

```bash
printf '\nexport const a = (  n:number )=>{return n*2}\n' >> apps/frontend/src/utils/number-formatter.ts
git commit -am "test"
git push    # fails
git push    # fails identically, and will keep doing so
```

### Suggested fix — one word

```diff
-        entry: npx prettier --write
+        entry: npx prettier --check
```

`--check` only reports which files need formatting and never edits them, so it can't create the uncommitted edit that causes this. You get a message naming the file, fix it with `npm run format`, and push once more.

Worth noting this may be part of what the commit message describes as the hooks being "very flaky and most developers skipped them" — a hook that dead-ends your push is a strong reason to reach for `--no-verify`, and once you do that you skip all of them.

.pre-commit-config.yaml
```diff
@@ -2,13 +2,6 @@ default_install_hook_types: [pre-push]
...
-      - id: type-check
-        name: type-check
-        entry: npm run type-check
-        language: system
-        stages: [pre-push]
-        fail_fast: true
-        pass_filenames: false
       - id: format
         name: format
         entry: npx prettier --write
```
bfliao, line 7: Preferrably we want: entry: **npx prettier --check**. There should be a comment stating why. Tldr: it's a livelock between prettier and pre-commit.

